PR #93, opened 2015-02-11
Expose version and revision

### LINKED ISSUE #91: Expose the editable.js version

peyerluk:
Fixes #91


peyerluk:
Since we don't use browserify in this project we need to replace the version and revision in the editable.js file. 

@marcbachmann maybe you have an idea how to do this with as little effort as possible. (I added the version.json like in the other projects but eventually that's not even needed.)

PR #99, opened 2015-03-12
Prevent illegal markup

peyerluk:
- Editable should not set links if the host or a parent of the host is a link tag
- Consider adding general rules for allowed elements

### DIFF
--- a/src/content.js
+++ b/src/content.js
@@ -186,6 +186,19 @@ var content = (function() {
       }
     },
 
+    getParentsAndSelf: function(node, tagName) {
+      var parents = [];
+      while (node) {
+        if (node.nodeType === nodeType.elementNode) {
+          if (!tagName || node.nodeName === tagName.toUpperCase()) {
+            parents.push(node);
+          }
+        }
+        node = node.parentNode;
+      }
+      return parents;
+    },
+
     /**
      * Get all tags that start or end inside the range
      */
--- a/src/selection.js
+++ b/src/selection.js
@@ -71,6 +71,16 @@ var Selection = (function() {
     },
 
     /**
+     * Check if a link can be added.
+     * If the editableHost or a parent is a link element linking
+     * should not be possible.
+     */
+    canLinkBeSet: function() {
+      var linkParents = content.getParentsAndSelf(this.host, 'a');
+      return linkParents.length ? false : true;
+    },
+
+    /**
      *
      * @method link
      */


src/selection.js
```diff
@@ -71,6 +71,16 @@ var Selection = (function() {
     },
 
     /**
+     * Check if a link can be added.
+     * If the editableHost or a parent is a link element linking
+     * should not be possible.
+     */
+    canLinkBeSet: function() {
+      var linkParents = content.getParentsAndSelf(this.host, 'a');
```
marcbachmann: @peyerluk This iterates throught all parents of the dom up to the html tag.
I guess we shouldn't go further than the container in which editable.js is initialized.

peyerluk: @marcbachmann This is on purpose since nesting links should not be allowed in general. Especially in case of teasers the link tag can be outside of the editable.

marcbachmann: Ok. It's still forbidden in HTML5 :)
It's not allowed to place interactive elements inside an `<a>` element:
http://www.w3.org/html/wg/drafts/html/master/single-page.html#the-a-element
http://www.w3.org/html/wg/drafts/html/master/single-page.html#interactive-content-2

src/content.js
```diff
@@ -186,6 +186,19 @@ var content = (function() {
...
+      while (node) {
+        if (node.nodeType === nodeType.elementNode) {
+          if (!tagName || node.nodeName === tagName.toUpperCase()) {
+            parents.push(node);
+          }
+        }
+        node = node.parentNode;
+      }
+      return parents;
+    },
```
jenstroeger, line 200: @peyerluk, would it be possible to merge only this function if you don’t want to merge the patch below? I ask because it comes in handy in other contexts as well, e.g. avoiding `<em>` inside of a `<strong>` or others. 

To address @marcbachmann’s [concern](https://github.com/livingdocsIO/editable.js/pull/99/files#r26464779) perhaps a flag that stops traversal at a given node would be an agreeable compromise?

Lastly, I’d probably rename the function to `getAncestorsAndSelf()` to be more in line with the XPath axis [ancestor-or-self](https://developer.mozilla.org/en-US/docs/Web/XPath/Axes/ancestor-or-self).

peyerluk: This function is a bit of an odd duck currently as this is the only method that is meant to collect info outside of the host element.

It would need some polishing before we can merge it.

Would this method here help your use case? Or are you more interested in knowing which tags affect a cursor or selection within the host? For that we have the method `getTags()` in this file.

jenstroeger: > Would this method here help your use case?

@peyerluk, I think I used `getTagNames()` from PR https://github.com/livingdocsIO/editable.js/pull/198 in its stead.

PR #100, opened 2015-03-12
Change js lib paths

wwjhu:
Out of the box the examples don't work due to the paths to the libs in the index.html that are not correct.

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -239,11 +239,11 @@ editable.cursor(function(elem, cursor) {
 
     <!-- Scripts -->
 
-    <script src="jquery/jquery.js"></script>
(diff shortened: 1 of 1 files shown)


peyerluk:
Thanks. I will have to look into the gruntfile and check if we have to change some things there as well. Until now we were using `grunt dev` to start a local server that would serve those files. But it is probably unnecessary. And the changes you proposed are already implemented in the `gh-pages` branch.

peyerluk:
This is currently incompatible with the way we load editable.js from the `.tmp` folder when developing. We could move the index.html file directly into the project folder but as of now I'd rather not do this to keep things a bit more separated.

PR #107, opened 2015-08-19
NPM-ification. Rangy 1.3

lukasbuenger:
Hi everybody,

couple of things come with this pull request and I'm aware of the fact that certain things are very much opinionated. Just take a look at it and cherry-pick whatever you consider fitting your masterplan.

#### NPM-ification

When I installed the `upfront-editable` package from npm, I noticed that there are quite a few things missing that would allow seamless installation and use of editable.js in an npm/browserify/webpack environment. In order to have better support for these use cases, I added/changed the following:

* Added jQuery and rangy (1.3, more on that later) to the dependencies. 
* Added require-statements to all source files that are depending on either rangy or jQuery or both. I think with this we finally got rid of all the globals.
* Added build scripts to npm, that use [browserify-shim](https://github.com/thlorenz/browserify-shim) to inject rangy and jQuery globals in order to not break the current behavior of the lib but still make it possible for npm users to install and compile seamlessly. I really tried to make this work with grunt but gave up after an hour or so. I wasn't the first one [as I found out later](http://stackoverflow.com/questions/23019194/difficulty-getting-browserify-shim-working-with-grunt-browserify-2-0-2-as-tr). 
* Added `core.js` as entry file and now you can:
```js
var Editable = require('upfront-editable');
```

#### Rangy 1.3

When I first installed rangy with npm, I didn't realize that your repo still works with 1.2.3. 
The only thing that was broken was the `range-save-restore.spec.js` because [rangy handles the conversion of a range to HTML differently than it used to](https://github.com/timdown/rangy/blob/master/src/core/domrange.js#L515). All the other specs passed. If you consider your test coverage adequate, you should be fine with 1.3. 

That should be it. What do you think on that?


peyerluk:
:heart_eyes: thx. will check it out.

peyerluk:
I'm using the alias option for the shim in grunt in the `livingdocs-framework`:
```coffee
browserify:
      options:
        browserifyOptions:
          extensions: ['.coffee']
        transform: ['coffeeify']
        alias: ['./src/modules/jquery_browserify.coffee:jquery']
```

and the shim in `./src/modules/jquery_browserify.coffee`
```coffee
module.exports = $
```

peyerluk:
Regarding rangy 1.3 I just didn't had the time to give it a thorough testing so far. There are quite a few cases that are not tested in the unit tests because controling cursor, selection and focus programatically is often not the same as if a user had done the respective actions in the browser. So manual testing is necessary. 
But I do not expect many problems if at all.

lukasbuenger:
Shims are in line with your way of things now. 
When I'm running `karma:build` I get a couple of errors though, all because of rangy's [`getNativeRange`](https://github.com/timdown/rangy/blob/fd74190a43f34e0960182657e7aadc3aaa9612bf/src/core/wrappedselection.js#L252) returning undefined. I haven't seen this while clicking through the example et al.

lukasbuenger:
The errors I had were due to a bug in the alias section of the browserify config in `Gruntfile.js`.
Updated and tested all dependencies as well.

marcbachmann:
@peyerluk btw aliases in browserify are deprecated.

boennemann:
What is the status on this one? Would love to be able to install/require this via npm :)

peyerluk:
It's not tested yet and some dependencies should get updated. But that should be about it.

PR #128, opened 2016-04-19
Improve build/test, rewrite ES2015

### LINKED ISSUE #62: Check out Travis CI
Once we will go public we could think about having a continuous integration testing platform like https://travis-ci.org/ that is free for OS projects

### LINKED ISSUE #78: Update to rangy 1.3
Currently (August 2014) rangy is under heavy development. Once that is finished and hopefully 1.3 is released update to the newest version. (https://github.com/timdown/rangy)

### LINKED ISSUE #103: Exposing editable.js as AMD or CJS Modules
Hey,
Does you plan to exposes ```editable.js```  as AMD or CJS Modules ?

When I'm tryin to use ```editable.js``` in RequireJS, I got ```Uncaught ReferenceError: bowser is not defined``` on [line 5781 of editable.js](https://github.com/upfrontIO/editable.js/blob/19d3aabf3a28666cc5558eb374de12422d579a43/editable.js#L5781).
Maybe I forgot something in my require.config ? : 
```js
require.config({
    ...
    paths: {
        ...
        "editable": "bower_components/editable/editable"
    },
    shim: {
        ...
        'editable': {
            deps: ['jquery'],
            exports: 'editable'
        }
    }
});
```

### LINKED ISSUE #126: Configure codeclimate
Create a custom configuration for codeclimate for this repo.

Files that should be ignored:

- editable.js
- editable.min.js

boennemann:
- [x] remove grunt
- [x] remove bower
- [x] build with webpack Closes #103
- [x] make `require`able Closes #103, Closes #107, Closes #78
- [x] update to rangy 1.3 Closes #107
- [x] test with karma-wepack
- [x] use standard style
- [x] use CI Closes #62 
- [x] use Hound Closes #126
- [x] auto-deploy examples to gh-pages
- [x] use semantic-release
- [x] update readme.md
- [x] setup es6 toolchain
- [x] use es6 in src
- [x] use es6 in tests
- [x] use es6 in examples
- [x] rebase and add `BREAKING CHANGE:` information where required
- [ ] enable greenkeeper once merged


boennemann:
@peyerluk @masone This is now ready for review. I suggest using the `w=1` url attribute to review the es2015 refactoring commits, e.g. https://github.com/upfrontIO/editable.js/pull/128/commits/220ef9a817626c388dccf08d299a2e0bc25d3bf6?w=1. It hides whitespaces changes, so the diffs get meaningful again.

examples/index.css
```diff
@@ -1,3 +1,7 @@
+@import "~normalize.css";
```
masone: what is that tilde in the @import doing?

masone:
This is pretty hard to review in detail, since there is so much new goodness!! Thanks for... webpack, es6, semantic release, greenkeeper, etc. This is a big step towards our desired setup for all of our projects.

I don't know much about ES6 yet, but it looks reasonable to me. I looked over the rest as well and if the tests run and you gave the new build a test run in our editor, I would go for a merge without picking too much on details.

examples/index.css
marcbachmann: normally `@import` imports a local file. When you prefix it with `~`, it loads the file from a node module.
So basically it's doing `@import(${require.resolve('normalize.css')})`. `normalize.css` is a node module.

masone: 🍭 sweet thanks

peyerluk:
Yes! A lot of goodness indeed :) 

How did you convert the coffee code by hand or did you use a tool?

examples/index.css
boennemann: CSS imports don't differentiate between `./foo.css` and `foo.css` (both relative), which is why for webpack it becamse kinda standard to use the tilde prefix to signify the `node_modules` folder. Which means this directly imports the `normalize.css` package's `main` file, which happens to be correctly set to the libraries' css file :)

peyerluk:
Ups, I forgot this was never a ☕ project... *duh*

peyerluk:
Very nice! Checked it out locally and it looks great :) 🚀

src/content.js
```diff
@@ -1,433 +1,382 @@
...
+ * Remove elements that were inserted for internal or user interface purposes
+ *
+ * @param {DOM node}
+ * @param {Boolean} whether to keep ui elements like spellchecking highlights
+ * Currently:
+ * - Saved ranges
+ */
+export function unwrapInternalNodes (sibling, keepUiElements) {
+  while (sibling) {
+    const nextSibling = sibling.nextSibling
```
DaRaFF: Why do you declare this as const? Is it a good idea to define everything as const, as soon as you know it gets not changed? I just want to point out, that `nextSibling` doesn't look like the typical const type like f.e. in Java. I just mention this, because I see you're using const a lot in this PR and want to find out, when you have planned to use const.

marcbachmann: > Constants are block-scoped, much like variables defined using the let statement. The value of a constant cannot change through re-assignment, and it can't be redeclared.

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Statements/const

karma.conf.js
```diff
@@ -1,77 +1,43 @@
...
-// Most of the options can be overriden by cli arguments (see karma --help)
-//
-// For all available config options and default values, see:
-// https://github.com/karma-runner/karma/blob/stable/lib/config.js#L54
-module.exports = function(config) {
-  config.set({
+process.env.BUILD_TEST = 'true'
+var webpackConfig = require('./webpack.config')
+delete webpackConfig.entry
+delete webpackConfig.output
```
marcbachmann: Can you use `_.omit(webpackConfig, 'entry', 'output')` here?
Otherwise this might have side effects once we restructure something.

masone:
There is one last comment from Marc to tackle. Let's get this merged and sort out the details on the go!

karma.conf.js
marcbachmann: We don't use underscore nor lodash in this project. It's more complicated than that.
So we probably have to use something like that.
`webpackConfig = Object.assign({}, webpackConfig, {entry: undefined, output: undefined})`

boennemann: Can you elaborate what the side effects are? Because it's modifying the original object?

marcbachmann: > Because it's modifying the original object?

yes. Karma is running in the same node process as webpack and therefore everything you do without a restart uses that config config object.

boennemann: There is at least one process for every `npm run` command, so I don't see where this could conflict. Inside of karma I'm deleting these entries precisely because I don't want webpack to pick them up + the config is passed explicitly.

When running `"test": "webpack && karma"` for example these would be two different processes.

src/content.js
boennemann: This is a bit of a stylistic question whether to do this or not, but it's something I've carried over from functional programming.
For classic constants I'd still use the uppercase convention e.g. `const REFRESH_INTERVALL = 100`, but it doesn't hurt to use const for regular variables as well.
Code that doesn't modify variables generally tends to have less side effects, to run faster and to be more readable. In this case `const` is simply used as the default variable type so redeclarations can never happen accidentally or unintentionally. If you need to modify one at a later point you're always free to change things to `let` or `var`, but then your making a conscious decision, otherwise the compiler will warn you.
It's meant to nudge everyone towards best practices.

masone: 👍

karma.conf.js
marcbachmann: > When running "test": "webpack && karma" for example these would be two different processes.
 
As long as we don't use the karma api, or a module that uses it, that's fine.
IMO we can merge this. We don't work that much in this project. So we probably won't run into problems the next year.

boennemann:
This will be released as 1.0.0 now, so nothing will break without explicitly updating.
I've linked this against Livingdocs – still works.

Got a few +1s, merging now, let's address things in other PRs from now on.

:shipit: 🎉

masone:
Congrats! 💯

PR #136, opened 2016-05-19
Enable `switch` event support

Fedia:
Allow `obj.on('event1 event2', function(){ ... })`


peyerluk:
I think this would be nice if we would pass an `event` object to listeners so they could figure out what type of event is being fired.

The way the events work now I fear this feature could lead to confusion some events have similar params and others different ones. I'd rather keep it simple for now.

Thanks for the contribution and sorry for answering so late!

PR #138, opened 2016-06-22
fix(eventable): don't execute events in wrong order on every 2nd call

boennemann:


boennemann:
Added (at first) failing regression test.

spec/eventable.spec.js
```diff
@@ -131,4 +131,31 @@ describe('eventable', () => {
       })
     })
   })
+
+  describe('notify()', () => {
+    let results, obj
+
```
peyerluk: The initialisation with `0` irritated me bit when reading the test.

spec/eventable.spec.js
```diff
@@ -131,4 +131,31 @@ describe('eventable', () => {
...
+    beforeEach(() => {
+      results = [0]
```
peyerluk: The initialisation with `0` irritated me bit when reading the test.

src/eventable.js
```diff
@@ -82,11 +83,13 @@ function getEventableModule (notifyContext) {
       }
 
       const eventListeners = listeners[event]
-      if (eventListeners === undefined) return
+      if (!Array.isArray(eventListeners)) return
```
peyerluk: Not sure about this more explicit test. If eventListeners is another type this will now silently return. I think it would be better if it threw an exception as any other type than undefined would be an error in `eventable`.

spec/eventable.spec.js
boennemann: Yeah true, not necessary.

src/eventable.js
```diff
@@ -82,11 +83,13 @@ function getEventableModule (notifyContext) {
...
 
       // Traverse backwards and execute the newest listeners first.
       // Stop if a listener returns false.
-      eventListeners.reverse().every((listener) => listener.apply(context, args) !== false)
+      eventListeners
+      .slice().reverse() // reverse is also in-place, so copy array before
```
peyerluk: the 'also' is unnecessary, or what are you referring to?

boennemann: The thing that's confusing me with reverse is that it's modifying the array in-place and also returning the reference.

Other in-place methods usually return something different: e.g. `[].push` returns the new index, `[].splice` returns the elements that it removed etc.

If `[].reverse` wouldn't return the reference this would've failed earlier. The way it's implemented it makes you believe it's not in-place.

src/eventable.js
boennemann: The explicit check for `undefined` isn't really necessary, because `if (!eventListeners)` would already suffice. Instead of just shortening it down I wanted to make the code more human readable.

src/eventable.js
boennemann: Or to sum it up in a single code snippet: 

```js
array === array.reverse()
// true
```

PR #140, opened 2016-10-03
Update dependencies

masone:


masone:
@peyerluk care to have a look?

masone:
I get a `events.js:321 Uncaught ReferenceError: React is not defined` when removing the `React` from the `import` statement

peyerluk:
@masone Thanks for updating the dependencies. How can I help? Should I test it, or do I have to look into something?

masone:
Yes please! I don't know how to handle the error stated in the comment above. The linter asks me to remove the unused variable `React`

```
  2:8  error  'React' is defined but never used  no-unused-vars
```

but it breaks when I remove it. Any idea what's going on? Tests and demo work, it's just the linter who's unhappy

peyerluk:
I guess that is reacts jsx that will be transformed into `React.createElement(...)` commands. Not sure why jslint complains about it. There are some react settings defined in `.eslintrc` but I don't know how they work in detail.

I'll have a look.

masone:
ah that makes sense! maybe we can set up some ignores for eslint?

masone:
was able to fix it, ready for review!

masone:
Totally forgot about this one. Thanks for the fix 😁

PR #145, opened 2017-02-27
allow to set attributes on bold and italic

gabrielhase:
@peyerluk This is just a rough proposal. It was quicker doing it in code 😄 
Give me a go if you're happy with the concept and I'll implement this for real.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -10,15 +10,26 @@ export const editableClass = 'js-editable'
 export const editableDisabledClass = 'js-editable-disabled'
 export const pastingAttribute = 'data-editable-is-pasting'
 export const boldMarkup = {
-  tagName: 'strong',
-  attributes: [
-    {
-      'attributeName': 'class',
-      'attributeValue': 'foo'
-    }
-  ]
+  type: 'tag',
+  name: 'strong',
+  attribs: {
+    'class': 'foo'
+  }
 }
-export const italicTag = 'em'
+export const italicMarkup = {
+  type: 'tag',
+  name: 'em',
+  attribs: {
+    'class': 'bar'
+  }
+}
+export const linkMarkup = {
+  type: 'tag',
+  name: 'a',
+  attribs: {
+    'class': 'foo bar'
+  }
+}
 
 // Rules that are applied when filtering pasted content
 export const pastedHtmlRules = {
--- a/src/config.js
+++ b/src/config.js
@@ -9,7 +9,15 @@ export const logErrors = true
 export const editableClass = 'js-editable'
 export const editableDisabledClass = 'js-editable-disabled'
 export const pastingAttribute = 'data-editable-is-pasting'
-export const boldTag = 'strong'
+export const boldMarkup = {
+  tagName: 'strong',
+  attributes: [
+    {
+      'attributeName': 'class',
+      'attributeValue': 'foo'
+    }
+  ]
+}
 export const italicTag = 'em'
 
 // Rules that are applied when filtering pasted content
--- a/src/config.js
+++ b/src/config.js
@@ -12,23 +12,17 @@ export const pastingAttribute = 'data-editable-is-pasting'
 export const boldMarkup = {
   type: 'tag',
   name: 'strong',
-  attribs: {
-    'class': 'foo'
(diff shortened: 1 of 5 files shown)


src/cursor.js
```diff
@@ -237,8 +237,12 @@ export default class Cursor {
 
   // Create an element with the correct ownerWindow
   // (see: http://www.w3.org/DOM/faq.html#ownerdoc)
-  createElement (tagName) {
-    return this.win.document.createElement(tagName)
+  createElement (tagName, attributes = []) {
+    const element = this.win.document.createElement(tagName)
+    for (const {attributeName, attributeValue} of attributes) {
```
marcbachmann: I think that an object for the attributes is more common.
It's closer to the dom api. I'd prefer that. `attributes: {class: 'foo'}`

gabrielhase:
@peyerluk I need your feedback here to proceed.

src/config.js
```diff
@@ -9,7 +9,15 @@ export const logErrors = true
 export const editableClass = 'js-editable'
 export const editableDisabledClass = 'js-editable-disabled'
 export const pastingAttribute = 'data-editable-is-pasting'
-export const boldTag = 'strong'
+export const boldMarkup = {
```
peyerluk: We can use the same syntax as in the framework:
```js
export const boldMarkup = {
  type 'tag'
  name: 'strong'
  attribs: {
    'class': 'foo'
  }
}
```

src/cursor.js
peyerluk: 👍 and see my comment above in the config.js

src/config.js
gabrielhase: 👍  will do

src/cursor.js
```diff
@@ -239,7 +239,9 @@ export default class Cursor {
   // (see: http://www.w3.org/DOM/faq.html#ownerdoc)
   createElement (tagName, attributes = []) {
     const element = this.win.document.createElement(tagName)
-    for (const {attributeName, attributeValue} of attributes) {
+    // I'd like to use for .. of with Object.entries() here but no IE support...
+    for (const attributeName of Object.keys(attributes)) {
```
marcbachmann: Can't you just use `for key in object`?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

```
for (const key in attributes) element.setAttribute(key, attributes[key])
```

gabrielhase: I could, but do you really think that is nicer?

marcbachmann: definitely. you can keep the braces and newlines if you like 😉

marcbachmann: And please fix the fallback
```
createElement (tagName, attributes = {}) {...
```

gabrielhase: good catch, thx!

gabrielhase:
@peyerluk please check again.

src/config.js
```diff
@@ -9,8 +9,27 @@ export const logErrors = true
 export const editableClass = 'js-editable'
 export const editableDisabledClass = 'js-editable-disabled'
 export const pastingAttribute = 'data-editable-is-pasting'
-export const boldTag = 'strong'
-export const italicTag = 'em'
+export const boldMarkup = {
+  type: 'tag',
+  name: 'strong',
+  attribs: {
+    'class': 'foo'
```
peyerluk: Please remove

src/config.js
```diff
@@ -9,8 +9,27 @@ export const logErrors = true
...
+  }
+}
+export const italicMarkup = {
+  type: 'tag',
+  name: 'em',
+  attribs: {
+    'class': 'bar'
```
peyerluk: same here

src/config.js
```diff
@@ -9,8 +9,27 @@ export const logErrors = true
...
+  }
+}
+export const linkMarkup = {
+  type: 'tag',
+  name: 'a',
+  attribs: {
+    'class': 'foo bar'
```
peyerluk: and here

gabrielhase: really, this one too? But it's `foo bar` 😆

PR #146, opened 2017-05-17
Normalize newlines on paste

caillou:
The browser represents newlines in HTML as spaces, and collapses multiples spaces.

This PR normalizes pasted content. The aim being for the HTML to be as close as possible to its representation in the browser.


caillou:
@peyerluk could you review this? This should generate a new patch version.

PR #148, opened 2017-10-31
fix(clipboard): Always get paste clipboard data as text/plain

vhf:
Tested against:

* Chrome 61.0
* Firefox 56.0
* Safari 11.0


src/dispatcher.js
```diff
@@ -95,6 +95,10 @@ export default class Dispatcher {
     .on('paste.editable', selector, function (event) {
       const element = this
 
+      event.preventDefault()
+      const textData = event.originalEvent.clipboardData.getData('text/plain')
+      self.$document.get()[0].execCommand('insertHTML', false, textData)
```
marcbachmann: doesn't `self.$document.get(0)`  do the same as `.get()[0]`?

vhf: Ah, yes, right, thanks!

vhf:
Thanks! (Yep, sometimes looking at the right place helps a lot! 😉 )

peyerluk:
Can you please describe the issue?

We have an existing past logic that parses HTML. So the title of this PR does not sound to be compatible with the existing logic as is. But I would like to hear the details.

vhf:
I don't have more details than what I got from @gabrielhase , let's ask him instead. 👍

src/clipboard.js
```diff
@@ -4,7 +4,8 @@ import * as config from './config'
 import * as string from './util/string'
 import * as nodeType from './node-type'
 
-let allowedElements, requiredAttributes, transformElements, blockLevelElements, splitIntoBlocks
+let allowedElements, requiredAttributes, transformElements, blockLevelElements
```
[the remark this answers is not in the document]
marcbachmann: weird, the eslint config in this project is completely outdated.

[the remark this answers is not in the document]
marcbachmann: and we use `"extends": "standard"`, which doesn't use semicolons.

vhf:
@peyerluk Up for review! :)

spec/clipboard.spec.js
```diff
@@ -129,5 +129,16 @@ describe('Clipboard', () => {
 
       expect(parseContent(div)[0]).toEqual('&lt;b&gt;a&lt;/b&gt;')
     })
+
+    it('removes blacklisted HTML elements (e.g. <style>)', () => {
+      const div = document.createElement('div')
+      div.innerHTML = `<style type="text/css">
+        p.p1 {font: 30.0px Georgia; color: #000000; -webkit-text-stroke: #000000}
```
peyerluk: I'm a fan of minimal tests. You could simplify this one a bit.

E.g
```js
div.innerHTML = `<style type="text/css">
    .foo { color: red; }
   </style>
   <p class="foo">
     bar
   </p>`

expect(parseContent(div)[0]).toEqual('bar')
```

src/config.js
```diff
@@ -63,5 +63,7 @@ export const pastedHtmlRules = {
   blockLevelElements: [
     'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'div', 'p', 'pre', 'hr', 'blockquote',
     'article', 'figure', 'header', 'footer', 'ul', 'ol', 'li', 'section', 'table', 'video'
-  ]
+  ],
+
+  blacklistedElements: ['style']
```
peyerluk: Please add a comment here describing what this does.

peyerluk:
@vhf Do you have an idea why hound is acting weird? Is the `.eslintrc` format we use outdated?

spec/clipboard.spec.js
vhf: Fixed!

src/config.js
vhf: Done, I added another comment where I thought it was relevant, please confirm its accuracy.

vhf:
`.eslintrc` looks fine to me, it also works well locally. I added a commit with a `.hound.yml`, it might get better. Not sure how to reproduce though, all I can do is hope it won't happen again on future PRs, I guess.

PR #150, opened 2017-11-10
Fix paste behaviour

mthpvg:
# Planning
https://github.com/upfrontIO/livingdocs-planning/issues/1576

## To reproduce

1. Make a very long paragraph that is longer than one viewport height (should have a scrollbar)
2. Move to the end of that paragraph and copy some content to the end

## Observed Behavior

The scroll position scrolls to the top of the paragraph and the place where the content was pasted is not visible anymore.

## Expected Behavior

The scroll position should not change.

## Video

https://drive.google.com/open?id=0Bya_F0ISZ6F8OUV4TTUwMmxNTU0

# Description
- The first commit fixes some import issues.
- The second commit tackles the main issue from the planning. We now disable the focus if we paste only one paragraph.

@peyerluk it looked to me as the simplest fix for the issue at hand. Can you assess its relevance? I'll look if it needs testing afterward.

### DIFF
--- a/src/cursor.js
+++ b/src/cursor.js
@@ -5,7 +5,7 @@ import * as content from './content'
 import * as parser from './parser'
 import * as string from './util/string'
 import * as nodeType from './node-type'
-import * as error from './util/error'
+import error from './util/error'
 import * as rangeSaveRestore from './range-save-restore'
 
 /**
--- a/src/cursor.js
+++ b/src/cursor.js
@@ -92,8 +92,21 @@ export default class Cursor {
     // Without setting focus() Firefox is not happy (seems setting a selection is not enough.
     // Probably because Firefox can handle multiple selections).
     if (this.win.document.activeElement !== this.host) {
+      const previousWindowTop = $(window).scrollTop()
+
       $(this.host).focus()
+
+      const hostBottom = $(this.host).position().top + $(this.host).outerHeight()
+      const windowHeight = $(window).height()
+      const isHostBottomVisible = hostBottom < previousWindowTop + windowHeight && hostBottom > previousWindowTop
+
+      if (isHostBottomVisible) {
+        $(window).scrollTop(previousWindowTop)
+      } else {
+        $(window).scrollTop(parseInt(hostBottom - windowHeight / 2, 10))
+      }
     }
+
     rangy.getSelection(this.win).setSingleRange(this.range)
   }
 
--- a/src/range-save-restore.js
+++ b/src/range-save-restore.js
@@ -1,5 +1,5 @@
 import rangy from 'rangy'
-import * as error from './util/error'
+import error from './util/error'
 import * as nodeType from './node-type'
 
 /**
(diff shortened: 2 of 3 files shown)


peyerluk:
Have you seen the comment in the code? The setting of the focus is there for Firefox.

And what is the reasoning behind only disabling the focus when one paragraph is pasted?

peyerluk:
It might be easier to just preserve the scroll position when focusing the element. I can help with that if you want.

mthpvg:
> Have you seen the comment in the code? The setting of the focus is there for Firefox.

In Chrome, the setting of the focus is the issue, it sends you back to the top of the paragraph even if the paragraph is really long.

In Firefox, it does nothing, we stay at the bottom of our long paragraph. In a sense, the issue is not present in Firefox.

> It might be easier to just preserve the scroll position when focusing the element. I can help with that if you want.

Good hint, I'll try that, and go back to you if needed, thanks.

mthpvg:
As you suggested: I now revert the scroll position when focusing.

Also, when the cursor is not visible anymore it sets the cursor's position at the center of the user's screen. The experience is nice IMHO, but I can remove it.

Before:
https://drive.google.com/open?id=1ApFx90fjKIAFWBct4Y-bK7BC1rIb9gMj

After:
https://drive.google.com/open?id=1gtvX3h_MD5Wa0Eg583i-yShGRYZn_Bzp

@peyerluk

src/cursor.js
```diff
@@ -92,8 +92,21 @@ export default class Cursor {
     // Without setting focus() Firefox is not happy (seems setting a selection is not enough.
     // Probably because Firefox can handle multiple selections).
     if (this.win.document.activeElement !== this.host) {
+      const previousWindowTop = $(window).scrollTop()
```
peyerluk: You must always use the `window` of the cursor and not the global `window`. Editable is set up to work with multiple iframes.

src/cursor.js
```diff
@@ -92,8 +92,21 @@ export default class Cursor {
...
+
       $(this.host).focus()
+
+      const hostBottom = $(this.host).position().top + $(this.host).outerHeight()
```
peyerluk: I like the idea but I think that might a bit too much for a low level library as editable. Plus I'm not sure if the pasted text is large if the cursor is still visible in all cases.

peyerluk: And here's an example that doesn't use jquery: https://github.com/upfrontIO/editable.js/commit/83eb19c7e4d3fb6c388d853c9ff0a1b5d5b50e7e

We try to move away from jQuery in editable. So I try to do new code without.

peyerluk: And we have existing code that reads the scroll position in this file. So we can reuse that logic.

mthpvg:
Closed in favor of: https://github.com/upfrontIO/editable.js/pull/151

src/cursor.js
mthpvg: Good to know.

src/cursor.js
mthpvg: 👍

PR #151, opened 2017-11-14
Preserve scroll position

mthpvg:
# Planning
https://github.com/upfrontIO/livingdocs-planning/issues/1576

# Description
This PR supersedes https://github.com/upfrontIO/editable.js/pull/150.

By just cancelling the scroll like we do, there is now an unexpected behavior. Before, when pasting multiple paragraphs it would focus on the last one. Now the view does not move.
As you said, this is a low level library and this behavior is not as counterintuitive as the one we are fixing here.

@peyerluk

### DIFF
--- a/src/cursor.js
+++ b/src/cursor.js
@@ -1,5 +1,6 @@
 import $ from 'jquery'
 import rangy from 'rangy'
+import viewport from './util/viewport'
 
 import * as content from './content'
 import * as parser from './parser'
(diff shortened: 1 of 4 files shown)


gabrielhase:
@mthpvg what is the status here? Nobody is assigned as a reviewer

mthpvg:
> @mthpvg what is the status here? Nobody is assigned as a reviewer

My bad, I mentioned @peyerluk in the post but I did not assign him as a reviewer.

gabrielhase:
@mthpvg in order to close the issue for real we need to release this up to the framework and editor / server. can you do that?

mthpvg:
> @mthpvg in order to close the issue for real we need to release this up to the framework and editor / server. can you do that?

Will do 👍 .

PR #163, opened 2019-02-19
Fix: Async ranges on focus

lukasbuenger:
Related to:
- livingdocsIO/livingdocs-planning#2559

#### Before:
When a `focus` event is not immediately followed by a selection event, the rangy `Range` object and the native `DOMRange` are out of sync, due to the browser not being able to recreate the previous selection after the sanitation that happens on `blur`.

#### Now:
When a `focus` event occurs, the rangy `Range` object gets synced with the `DOMRange` by calling the `select` behavior as well.
When losing and regaining focus to and from any event not related to selection, we just collapse to start.


spec/create-default-events.spec.js
```diff
@@ -0,0 +1,90 @@
+import $ from 'jquery'
+import rangy from 'rangy'
+
+import Cursor from '../src/cursor'
+import Editable from '../src/core'
+
+describe('Default Events', () => {
+  let $elem, editable, focus, blur
```
peyerluk: You should write these onto `this` in mocha. This helps keeping test scopes separated.

spec/create-default-events.spec.js
```diff
@@ -0,0 +1,90 @@
...
+  function off () {
+    if (onListener) {
+      editable.unload()
+      onListener = undefined
+    }
+  }
+
+  describe('for editable', () => {
+    beforeEach(() => {
+      $elem = $('<div contenteditable="true"></div>')
```
peyerluk: The `contenteditable` attribute will be added by `editable.add` you can omit it here.

src/dispatcher.js
```diff
@@ -84,7 +84,8 @@ export default class Dispatcher {
     this.$document
       .on('focus.editable', selector, function (event) {
         if (this.getAttribute(config.pastingAttribute)) return
-        self.notify('focus', this)
+        const selection = self.selectionWatcher.getFreshSelection()
```
peyerluk: Can we be sure that a selection is always within the current element?

src/create-default-events.js
```diff
@@ -11,8 +11,9 @@ export default function createDefaultEvents (editable) {
      * @event focus
      * @param {HTMLElement} element The element triggering the event.
      */
-    focus (element) {
+    focus (element, selection) {
       behavior.focus(element)
+      behavior.selection(element, selection)
```
peyerluk: Maybe we could add logic to determine if the `selection` has actually changed so we don't emit every selection event twice after this change.

What do you think?

lukasbuenger:
ToDo:

[ ] New function in `selection-watcher` like `syncSelection` or `updateSelection`.
[ ] Remove `behavior.select` call

PR #164, opened 2019-02-27
feat(highlighting): Can display textrange-based highlights

lukasbuenger:
Relations:
  - Issue: livingdocsIO/livingdocs-planning/issues/2472
  - Related PR's: 
     - livingdocsIO/livingdocs-framework#378
     - livingdocsIO/livingdocs-editor#2646

### Changelog

* API method `highlight` now takes an additional optional option argument `textRange {start: number, end: number, text: string }`. If given, it is used to determine the span position rather than the regex matches on `text`.
* Added rangy's [TextRange module](https://github.com/timdown/rangy/wiki/Text-Range-Module).

### Known issues

* Formatting a selection that includes a comment edge will remove any comment span in the editable container
* Splitting a comment destroys the highlight.
* Splitting a component before a comment destroys the highlight.
* Merging a component destroys all highlights in the second component.

## Pull Request Checklist
- [x] Tests


spec/highlighting.spec.js
```diff
@@ -54,4 +53,86 @@ describe('Highlighting', function () {
...
+        text: 'bari',
+        highlightId: 'myId',
+        textRange: { start: 3, end: 7 }
+      })
+      this.editable.highlight({
+        editableHost: this.$div[0],
+        text: 'Foobari',
+        highlightId: 'myId',
+        textRange: { start: 0, end: 7 }
+      })
```
lukasbuenger: Allow update/overwrite.

spec/highlighting.spec.js
```diff
@@ -54,4 +53,86 @@ describe('Highlighting', function () {
...
+      this.editable.highlight({
+        editableHost: this.$div[0],
+        text: 'bari',
+        highlightId: 'myId',
+        textRange: { foo: 3, bar: 7 }
+      })
+      const highlightSpan = this.$div.find('[data-word-id="myId"]')
+      expect(highlightSpan.length).toEqual(0)
+    })
+  })
```
lukasbuenger: - minimalistic ranges
- 0-1 positions
- containing / overlapping ranges
- succeeding ranges

peyerluk: more use cases:
end position
ranges with newlines within
two identical ranges
markup (<b> and all possible contains and intersections)
other highlights overlapps

lukasbuenger: - textrange export on selection

lukasbuenger:
Closed for now in favor of #166

src/core.js
```diff
@@ -332,19 +331,59 @@ const Editable = module.exports = class Editable {
...
+   * The markup used for the highlighting will be removed
+   * from the final content.
+   *
+   *
+   * @param  {Object} options
+   * @param  {DOMNode} options.editableHost
+   * @param  {String} options.text
+   * @param  {String} options.highlightId Added to the highlight markups in the property `data-word-id`
+   * @param  {Object} [options.textRange] An optional range which gets used to set the markers.
+   * @param  {Number} [options.textRange.start]
```
peyerluk: I think `start` and `end` are not optional if a `textRange` object is passed, right? So I think we can remove the square brackets around the param name here and in the next line.

spec/highlighting.spec.js
```diff
@@ -54,4 +76,237 @@ describe('Highlighting', function () {
     })
   })
 
+  describe('highlightSupport', () => {
+    beforeEach(() => {
+      setupHighlightEnv(this, 'People Make The <br> World Go Round')
```
peyerluk: We should replace all arrow functions in this file with regular functions if you use `this` here. Otherwise this code is quite misleading.

peyerluk: (We could also do this in a separate PR for all tests in editable.js)

spec/highlighting.spec.js
```diff
@@ -54,4 +76,237 @@ describe('Highlighting', function () {
     })
   })
 
+  describe('highlightSupport', () => {
```
peyerluk: All these tests never test if we actually to the right thing! You only check that the index positions are the same, we do not know that the right characters are selected.

src/core.js
```diff
@@ -332,19 +331,59 @@ const Editable = module.exports = class Editable {
...
+
+  /**
+   * Extracts positions of all DOMNodes that match `[data-word-id]`.
+   *
+   * Returns an object where the keys represent a highlight id and the value
+   * a text range object.
+   *
+   * @param  {Object} options
+   * @param  {DOMNode} options.editableHos
+   * @return {Object} ranges
```
peyerluk: Why not include the text in the ranges? Otherwise how do you get the text of a range?

spec/highlighting.spec.js
lukasbuenger: I would like to go with your second suggestion, as currently a *lot* of tests misbehave if you only just adjust the `beforeEach` scopings.

src/core.js
lukasbuenger: Done!

spec/highlighting.spec.js
peyerluk: > I would like to go with your second suggestion

That's ok

> as currently a lot of tests misbehave if you only just adjust the beforeEach scopings.

This is because you have to change all methods in the same file at once, otherwise you mix scopes which does not work.

lukasbuenger: I did replace all of them to the point where I had now arrow functions at all anymore (at least in this spec), but still all hell broke loose. I'd rather do this with a little more focus and time.

spec/highlighting.spec.js
lukasbuenger: Tests now check for the text equivalent of the ranges. See also comment below.

src/core.js
lukasbuenger: > Why not include the text in the ranges? 
Done. Good point. Will have to adjust specs in the framework as well. JFYI

src/highlighting.js
```diff
@@ -25,7 +25,7 @@ export default class Highlighting {
       // unbounce rate in ms before calling the spellcheck service after changes
       throttle: 1000,
       // remove highlights after a change if the cursor is inside a highlight
-      removeOnCorrection: true,
+      removeOnCorrection: false,
```
peyerluk: Why did this change?

spec/highlighting.spec.js
```diff
@@ -54,4 +84,400 @@ describe('Highlighting', function () {
...
+      this.highlightRange('first', 11, 22)
+      this.highlightRange('first', 8, 9)
+      const expectedRanges = {
+        first: {
+          text: 'a',
+          start: 8,
+          end: 9
+        }
+      }
+      const expectedHtml = this.formatHtml(`People M
```
lukasbuenger: Two closing spans. Why?

lukasbuenger: This was my typo. The test was still passing because of [this](https://github.com/livingdocsIO/editable.js/pull/164/files#diff-3990f96ac4ef1792695bda53e7fd3ecfR30)

PR #165, opened 2019-03-14
Migrate to DroneCI v1

marcbachmann:


marcbachmann:
The travis build here doesn't use a shrinkwrap, so we're good to go.

PR #166, opened 2019-03-15
WIP Feat: VirtualSpan

lukasbuenger:
Relations:
  - Issues: 
      - livingdocsIO/livingdocs-planning#2472
      - livingdocsIO/livingdocs-planning#2473
  - PR's:
      - livingdocsIO/livingdocs-framework#381 
      - livingdocsIO/livingdocs-editor#2639

### Changelog
* Added rangy's [TextRange module](https://github.com/timdown/rangy/wiki/Text-Range-Module), mainly because of its `toCharacterRange` method. Exposed as `Selection#getTextRange`.

##### `VirtualSpan`
A class that helps with handling partials in text, which have a visual representation but are not considered part of the actual document. It turned out to be independent of a running Editable instance, that's why I made it part of a new `contrib` namespace. 
Its first use case would be comments in documents.

It exposes two kind of methods:
1. Helpers to insert, remove and update the DOM representations of a given set of virtual spans. Operates on the DOM. (See livingdocsIO/livingdocs-editor#2639)
2. Helpers to remove  and reapply said virtual spans from a given content *string*. (See livingdocsIO/livingdocs-framework#381)

The idea is to use the first set of methods to attach and remove virtual spans to the DOM and the second set to guarantee data consistency on a model level. 

The main motivation behind not having those virtual spans have the `ui-unwrap` directive and consequentially being removed and reapplied on every update (as I tried here livingdocsIO/editable.js#164) is that those virtual spans somewhat have to live in each components markup. They are expected to behave like any other inline tag (eg `<strong>`) and treating them as some extraneous entity felt like preventing editable from being editable while reimplementing its core features for the most parts.

It's a little unorthodox, but it's playing to the general strengths of this lib in a way that no other solution could (at least those I was able to come up with). And it's very contained and can easily be added/removed when needed.   

### Known issues

* Formatting a selection that includes a comment edge will remove any virtual span in the editable container aka *nuke* issue

## Pull Request Checklist
- [ ] Tests

### DIFF
--- a/src/contrib/virtual-span.js
+++ b/src/contrib/virtual-span.js
@@ -1,0 +1,152 @@
+import $ from 'jquery'
+import rangy from 'rangy'
+import { unwrap, adoptElement } from './content'
+
+function defaultMarkerConstructor (id, namespace) {
+  const marker = $('span')[0]
+  return marker
+}
+
+const defaultOptions = {
+  namespace: 'vspan',
+  idAttribute: 'data-vspan-id',
+  namespaceAttribute: 'data-vspan-namespace',
+  editableNamespace: 'vspan',
+  markerConstructor: defaultMarkerConstructor,
+  win: undefined
+}
+
+class VirtualSpan {
+  constructor ({
+    namespace,
+    idAttribute,
+    namespaceAttribute,
+    editableNamespace,
+    markerConstructor,
+    win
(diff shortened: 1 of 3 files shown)


mokimo:
This is stale and outdated, for this reason, I'll close this.

PR #169, opened 2019-09-04
Clipboard: Copy&Paste keeps links relative

mokimo:
Relations:
  - Issue: https://github.com/livingdocsIO/livingdocs-planning/issues/2965
  - Issue2: https://github.com/livingdocsIO/livingdocs-planning/issues/3025
## Description
The clipboard turns relative links into absolute ones so:
`/internal/link` gets converted to `https://livingdocs.io/internal/link`

There is a new flag to turn off this behavior if the link was copied from the same context/host editable is running on.

So if a relative link was copied from `https://livingdocs.io/internal/link` it strips off the protocol, hostname and port number of a URL and turns it to `/internal/link`
If a relative link was copied from `https://somewebpage.com/internal/link` it will still convert it to an absolute one.

## Changelog
- new config option
```js
Editable.globalConfig({
 pastedHtmlRules: {
  keepInternalRelativeLinks: true || false // default false
 }
}
```

## Test Instructions
Create an article.
Add a text with a relative link like /link.
Copy the text and past to another component.
Check the link.

### DIFF
--- a/spec/clipboard.spec.js
+++ b/spec/clipboard.spec.js
@@ -1,8 +1,14 @@
 import { parseContent, updateConfig } from '../src/clipboard'
+import cloneDeep from 'lodash.clonedeep'
 import * as config from '../src/config'
 
 describe('Clipboard', () => {
   describe('parseContent()', () => {
+
+    afterEach(() => {
+      updateConfig(config)
+    })
+
     function extract (str) {
       const div = document.createElement('div')
       div.innerHTML = str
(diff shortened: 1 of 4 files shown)


spec/clipboard.spec.js
```diff
@@ -144,11 +144,16 @@ describe('Clipboard', () => {
       expect(parseContent(div)[0]).toEqual('bar')
     })
 
-    it('keeps internal links of a anchorTag with an href attribute', () => {
-      expect(extractSingleBlock('<a href="http://localhost:9876/test123">a</a>')).toEqual('<a href="http://localhost:9876/test123">a</a>')
+    it('changes absolute links to relative ones with the keepInternalRelativeLinks flag set to true', () => {
       config.pastedHtmlRules.keepInternalRelativeLinks = true
       updateConfig(config)
```
peyerluk: Each test should restore the config to the state before. Or is that already done somewhere?

spec/clipboard.spec.js
```diff
@@ -142,5 +143,17 @@ describe('Clipboard', () => {
 
       expect(parseContent(div)[0]).toEqual('bar')
     })
+
+    it('changes absolute links to relative ones with the keepInternalRelativeLinks flag set to true', () => {
+      config.pastedHtmlRules.keepInternalRelativeLinks = true
+      updateConfig(config)
+      expect(extractSingleBlock('<a href="http://localhost:9876/test123">a</a>')).toEqual('<a href="/test123">a</a>')
```
peyerluk: It would be better not to hardcode `localhost:9876` here.

spec/clipboard.spec.js
```diff
@@ -142,5 +143,17 @@ describe('Clipboard', () => {
 
       expect(parseContent(div)[0]).toEqual('bar')
     })
+
+    it('changes absolute links to relative ones with the keepInternalRelativeLinks flag set to true', () => {
```
peyerluk: Can you also add a test where nothing gets replaced?

```js
expect(extractSingleBlock('<a href="/test123">a</a>'))
  .toEqual('<a href="/test123">a</a>')
```

and maybe also:

```js
expect(extractSingleBlock('<a href="http://foo.com/test123">a</a>'))
  .toEqual('<a href="http://foo.com/test123">a</a>')
```

src/clipboard.js
```diff
@@ -87,14 +90,19 @@ export function filterHtmlElements (elem) {
       return ''
     }
 
+    // Keep internal relative links relative (on paste).
+    if (keepInternalRelativeLinks && child.nodeName === 'A' && child.href) {
+      const stripInternalHost = child.href.replace(window.location.origin, '')
```
peyerluk: I would use `getAttribute('href')` here. That sounds more consistent. And `child.href` always adds `window.location.origin` and then you strip it away again which is not really necessary.

spec/clipboard.spec.js
peyerluk: Oh and:

```js
expect(extractSingleBlock('<a href="">')).toEqual('<a href="">')
```

and

```js
```js
expect(extractSingleBlock('<a href>')).toEqual('<a href>')
```
```

Comment (mokimo) [reply spec/clipboard.spec.js]: Added and there were some previous tests for the last 2 cases the expected outcome is that it's removed
```js
expect(extractSingleBlock('<a href>')).toEqual('<a>')
```

src/clipboard.js
mokimo: I don't really get this comment, the sense of this PR is to add the option to remove/strip `window.location.origin` from the `href` attribute  give that it's an relative link

PR #170, opened 2019-10-03
Add helper to retrieve the editable configs

mokimo:
### Description
Instead of resorting to using a deep-require - a simple helper to retrieve the Editable standard configs.

### Changelog
- Adds `Editable.getGlobalConfig()`to retrieve the current state of the configs

### DIFF
--- a/spec/config.spec.js
+++ b/spec/config.spec.js
@@ -39,6 +39,18 @@ describe('Editable configuration', () => {
       Editable.globalConfig(originalConfig)
     })
 
+    it('retreives the default config', () => {
+      expect(originalConfig).toEqual(Editable.getGlobalConfig())
+    })
+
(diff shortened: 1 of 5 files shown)


spec/config.spec.js
```diff
@@ -39,6 +39,18 @@ describe('Editable configuration', () => {
...
+    it('retreives the default config', () => {
+      expect(originalConfig).toEqual(Editable.getGlobalConfig())
+    })
+
+    it('retrieves the default config after changes to the globalConfig', () => {
+      Editable.globalConfig({
+        editableClass: 'editable-instance'
+      })
+
+      expect(originalConfig).not.toEqual(Editable.getGlobalConfig())
```
mokimo: Ah spotted a bug, this is supposed to be the same

PR #171, opened 2019-10-10
update model on remove and add highlight

meinradandermatt:
Relations:
  - Backport: < add a backport link in both directions (original <-> backport) >
  - Issue: https://github.com/livingdocsIO/livingdocs-planning/issues/3088

## Description

If the highlighting is changing a change event should be sent. This is because some highlighting like e.g for comments is part of the content so changes should be treated the same as content changes.

With this change on the setVisibleSelection the focus shouldn't be set because otherwise on writing a comment the focus will be set on the component.

## Changelog

- Don't force focus on `setVisibleSelection()` anymore.
the reason is the spellcheck which use the retainVisibleCollection and come back after some time and set the focus again. We should split the spellcheck from the standard highlighting

- Update api endpoint behavior for highlights:
```js
// Now triggers change events
// Also there is a new options param `raiseEvents` to raise change events.
highlight ({editableHost, text, highlightId, textRange, raiseEvents}) {}

// Now triggers change events if options param raiseEvents is set to true
removeHighlight ({editableHost, highlightId, raiseEvents}) {}
```

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -356,9 +356,10 @@ const Editable = module.exports = class Editable {
    * @param  {Object} [options.textRange] An optional range which gets used to set the markers.
    * @param  {Number} options.textRange.start
    * @param  {Number} options.textRange.end
+   * @param  {Boolean} options.raiseEvents do throw change events
    * @return {Number} The text-based start offset of the newly applied highlight or `-1` if the range was considered invalid.
    */
-  highlight ({editableHost, text, highlightId, textRange}) {
+  highlight ({editableHost, text, highlightId, textRange, raiseEvents}) {
     if (!textRange) {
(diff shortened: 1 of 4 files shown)


src/core.js
```diff
@@ -374,7 +375,7 @@ const Editable = module.exports = class Editable {
       )
       return -1
     }
-    return highlightSupport.highlightRange(editableHost, highlightId, textRange.start, textRange.end)
+    return highlightSupport.highlightRange(editableHost, highlightId, textRange.start, textRange.end, silent ? undefined : this.dispatcher)
```
peyerluk: This line is too long

meinradandermatt: change logic if raiseEvents = true

PR #172, opened 2019-11-15
Shift enter at end of text adds data tag to the added <br> tag

mokimo:
### Description
We're fighting against default browser behavior here as adding a break-tag with shift+enter at the end of a text, will automatically create two break-tags, with one being un-deletable.

Related issue that tinymc has (They have a WYSIWYG-editor): https://github.com/tinymce/tinymce/issues/4856

My suggestion is to just intervene and disallow/remove break-tags at the end of a text. This is more transparent due to the visual feedback, vs hidden `<br>` tags in the document.

Furthermore, mark all other break-tags (`<br data-editable="newline">`), because this solution has some tradeoffs as well.

### Changelog
- Shift+enter at the end of a line will now add `<br data-editable="remove">` instead of `<br>`
- Shift+enter now marks all the break-tags `<br data-editable="newline">`

### Tradeoffs
- When creating a new line with shift+enter in the middle of a text, then deleting everything after the soft-break, will result in an undeletable break-tag again.
-> Those would be manually filterable `<br data-editable="newline">` scanning for that.

### DIFF
--- a/spec/cursor.spec.js
+++ b/spec/cursor.spec.js
@@ -53,6 +53,12 @@ describe('Cursor', function () {
       })
     })
 
+    describe('isAtEnd()', () => {
+      it('is true', () => {
+        expect(this.cursor.isAtTextEnd()).toBe(true)
+      })
+    })
+
     describe('isAtBeginning()', () => {
       it('is false', () => {
         expect(this.cursor.isAtBeginning()).toBe(false)
@@ -66,6 +72,7 @@ describe('Cursor', function () {
         // move the cursor so we can check the restore method.
         this.cursor.moveAtBeginning()
         expect(this.cursor.isAtBeginning()).toBe(true)
+        expect(this.cursor.isAtTextEnd()).toBe(false)
 
         this.cursor.restore()
         expect(this.cursor.isAtEnd()).toBe(true)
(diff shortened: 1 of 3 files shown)


spec/cursor.spec.js
```diff
@@ -53,6 +53,12 @@ describe('Cursor', function () {
       })
     })
 
+    describe('isAtEnd()', () => {
```
peyerluk: correction: `isAtTextEnd()`

spec/cursor.spec.js
```diff
@@ -53,6 +53,12 @@ describe('Cursor', function () {
...
+      it('is true', () => {
```
peyerluk: 'returns true when at text end'

mokimo:
Let's take it in before the major bump with jquery, hoping looking a bit more into the tests later on

marcbachmann:
I think I just found the original cause for this issue. It's not the enter which caused issues but the `br` placeholder we added during a focus event. With the solution in #234 we now have `\u0000` as default content, which then results with `\u0000` after the content.
```
focus
type foobar
element.textContent equals 'foobar\u0000'
```


But as we strip the null escape character away, this doesn't cause any side effect anymore.

PR #174, opened 2019-12-12
Bump dependencies

mokimo:
As a part of some good new-year resolutions, let's bump some crucial dependencies.
(We should upgrade webpack, karma, etc. in a seperate PR at some point)

Closes https://github.com/livingdocsIO/editable.js/pull/153

## Changelog
- Dependency bump
```
🔥 Breaking change 🔥 :
We dropped support for jquery 2.x.x.
Please upgrade to jquery 3.x.x to ensure editable is working as it is supposed to.
```


package.json
```diff
@@ -4,7 +4,7 @@
   "version": "0.0.0-development",
   "dependencies": {
     "babel-runtime": "^6.6.1",
-    "bowser": "1.4.6",
+    "bowser": "2.7.0",
```
marcbachmann: Didn't bowser do a breaking change?

package.json
```diff
@@ -17,36 +17,36 @@
...
-    "sinon": "^2.4.1",
-    "style-loader": "^0.13.1",
-    "url-loader": "^0.5.7",
+    "react-dom": "^16.12.0",
+    "prop-types": "^15.7.2",
+    "rimraf": "^3.0.0",
+    "sinon": "^7.4.1",
+    "style-loader": "^1.0.1",
+    "url-loader": "^2.1.0",
     "webpack": "^1.13.0",
```
marcbachmann: What about webpack? :trollface:

package.json
marcbachmann: In the editor at least we had to do a change: https://github.com/livingdocsIO/livingdocs-editor/commit/d9626e8ca8e06066d9459eb43f68d546a2dce9f8

mokimo: Ah that's indeed problematic, it doesn't error but causes some problems I'll fix that

package.json
mokimo: > (We should upgrade webpack, karma, etc. in a seperate PR at some point)

Mainly wanted to up jquery

package.json
```diff
@@ -17,36 +17,35 @@
...
+    "eslint-plugin-react": "^7.17.0",
     "eslint-plugin-standard": "^3.1.0",
-    "file-loader": "^0.9.0",
     "font-awesome": "^4.6.1",
     "gh-pages": "^0.11.0",
-    "jasmine-core": "^2.2.0",
-    "jquery": "^2.2.4",
-    "karma": "^1.3.0",
+    "jasmine-core": "^2.99.1",
+    "jquery": "^3.4.1",
```
marcbachmann: Please also update the peerDependency https://github.com/livingdocsIO/editable.js/pull/174/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L68
And then this could also be a breaking change... depending on what range we declare in there.
Do we drop support for jquery 2?

mokimo: Would that be a breaking change from the editable perspective? Or do peer-dependencies take either 3.x.x and then if that's not available they take what's available so 2.x.?

mokimo: Seems so, in this case I marked it as breaking change.

Jquery 2.x.x should still work, we just won't actively support it IMO.

Issue #161, opened 2018-12-16
Switch event not firing?


BryceBubbles:
The documented switch event doesn't seem to fire.
I think tab, shift+tab and arrows do their default actions.

All other documented events work as expected.

BryceBubbles:
Looks like the switch event is only dispatched if the cursor stays in the same position on arrow press.
https://github.com/livingdocsIO/editable.js/blob/c2107cc2182d54f64322c744d7853aec6301cc38/src/dispatcher.js#L170-L179
I'm guessing `newCursor.equals(cursor)` is always false for some reason.
I haven't been able to look into why this might be yet.

peyerluk:
This is fixed now.

PR #179, opened 2020-05-14
Allow listening multiple event types

joas8211:
Allow to use eventable.on to listen multiple event types with the
same listener. 

Keyboard.on is called with multiple event types in dispatcher.js, so this fixes issue #161.
https://github.com/livingdocsIO/editable.js/blob/4b264b8cc15af9160cfa32b8ed8a088aae164553/src/dispatcher.js#L207-L216

### DIFF
--- a/src/eventable.js
+++ b/src/eventable.js
@@ -54,7 +54,7 @@ function getEventableModule (notifyContext) {
   return {
     on (event, listener) {
       if (arguments.length === 2) {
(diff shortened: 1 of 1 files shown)


peyerluk:
Thanks (and sorry for the long wait).

Fixed here: https://github.com/livingdocsIO/editable.js/pull/191

PR #181, opened 2020-07-10
add custom elements and 

meinradandermatt:
# Description
Add the possibility to toggle Custom elements and to detect the wrapping elements by tagName and Attributes. On this way a span can be used twice with other attributes.

# Changelog
- toggleTag checks on the tag and attributes
- on a selection a custom element can be toggled

### DIFF
--- a/spec/content.spec.js
+++ b/spec/content.spec.js
@@ -169,6 +169,38 @@ describe('Content', () => {
     })
   })
 
+  describe('getTagsByNameAndAttributes()', () => {
+    let range
+    beforeEach(() => {
+      range = rangy.createRange()
+    })
+
+    it('filters tag with attributes match', () => {
+      const test = $('<div><span class="foo"><span class="test">a</span></span></div>')
+      range.setStart(test[0], 0)
+      range.setEnd(test[0], 1)
+      const tags = content.getTagsByNameAndAttributes(test[0], range, $('<span class="foo">')[0])
+      expect(content.getTagNames(tags)).toEqual(['SPAN'])
+    })
+
+    it('filters tag with attributes match', () => {
+      const test = $('<div><span class="foo"><span class="foo">a</span></span></div>')
+      range.setStart(test[0], 0)
+      range.setEnd(test[0], 1)
+      const tags = content.getTagsByNameAndAttributes(test[0], range, $('<span class="foo">')[0])
+      expect(content.getTagNames(tags)).toEqual(['SPAN', 'SPAN'])
+    })
+
+    it('filters inner tags', () => {
+      // <div>|<span class="foo"><span class="test">a</span></span>|</div>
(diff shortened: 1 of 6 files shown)


spec/content.spec.js
```diff
@@ -169,6 +169,38 @@ describe('Content', () => {
     })
   })
 
+  describe('getTagsByNameAndAttributes()', () => {
```
peyerluk, line 172: Please always use `function () {}` instead of `() => {` when using mocha.

spec/content.spec.js
```diff
@@ -169,6 +169,38 @@ describe('Content', () => {
...
+    let range
+    beforeEach(() => {
+      range = rangy.createRange()
```
peyerluk, line 175: And here you can use `this.range = ...`. That's the default of how to set a variable with mocha.

spec/content.spec.js
peyerluk: ... I see it's done like this in the whole file, but it is still wrong as it error prone and doesnt properly isolate tests when `() => {}` is used together with `this`. We should quickly fix the whole file.

src/content.js
```diff
@@ -180,6 +180,24 @@ export function getTagsByName (host, range, tagName) {
...
+  return getTags(host, range, (node) => {
+    return node.nodeName === elem.nodeName.toUpperCase() &&
+      areSameAttributes(node.attributes, elem.attributes)
+  })
+}
+
+export function areSameAttributes (attrs1, attrs2) {
+  if (attrs1.length !== attrs2.length) return false
+
+  for (var i = 0; i < attrs1.length; i++) {
```
peyerluk, line 193: `var` -> `let`

src/content.js
```diff
@@ -180,6 +180,24 @@ export function getTagsByName (host, range, tagName) {
   })
 }
 
+export function getTagsByNameAndAttributes (host, range, elem) {
+  return getTags(host, range, (node) => {
+    return node.nodeName === elem.nodeName.toUpperCase() &&
```
peyerluk, line 185: why is there only one `toUpperCase()?  in here?

src/content.js
```diff
@@ -301,6 +325,18 @@ export function nuke (host, range, tagName) {
   })
 }
 
+// Unwrap all tags this range is affected by.
+// Can also affect content outside of the range.
+export function nukeElem (host, range, node) {
+  getTags(host, range).forEach((elem) => {
+    if (elem.nodeName !== 'BR' && (!node ||
+        (elem.nodeName === node.nodeName.toUpperCase() &&
```
peyerluk, line 333: same here: why is `toUpperCase()` only used once (this also goes for the line directly above: why not there?)

meinradandermatt: because I just adapted the nuke () function. But I changed it.

PR #183, opened 2020-08-18
add subscript & superscript

TorstenDittmann:
- add sub- and superscript tags
- add tests

### DIFF
--- a/spec/selection.spec.js
+++ b/spec/selection.spec.js
@@ -218,6 +218,68 @@ describe('Selection', function () {
       })
     })
 
+    describe('superscript:', function () {
+      beforeEach(function () {
+        this.oldSuperscriptMarkup = config.superscriptMarkup
+        config.superscriptMarkup = {
+          type: 'tag',
+          name: 'sup',
+          attribs: {
+            'class': 'bar'
+          }
+        }
+      })
+
+      afterEach(function () {
+        config.superscriptMarkup = this.oldSuperscriptMarkup
+      })
+
+      it('makes the selection superscript with the configured class', function () {
+        this.selection.makeSuperscript()
+        const superscriptTags = this.selection.getTagsByName('sup')
+        const html = getHtml(superscriptTags[0])
+        expect(html).toEqual('<sup class="bar">foobar</sup>')
+      })
+
+      it('toggles the superscript selection', function () {
+        this.selection.makeSuperscript()
+        this.selection.toggleSuperscript()
+        const superscriptTags = this.selection.getTagsByName('sup')
+        expect(superscriptTags.length).toEqual(0)
+      })
+    })
+
+    describe('subscript:', function () {
+      beforeEach(function () {
+        this.oldSubscriptMarkup = config.subscriptMarkup
+        config.subscriptMarkup = {
+          type: 'tag',
+          name: 'sub',
+          attribs: {
+            'class': 'bar'
+          }
+        }
+      })
+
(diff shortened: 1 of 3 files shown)


spec/selection.spec.js
```diff
@@ -218,6 +218,68 @@ describe('Selection', function () {
...
     })
 
+    describe('superscript:', function () {
+      beforeEach(function () {
+        this.oldSuperscriptMarkup = config.superscriptMarkup
+        config.superscriptMarkup = {
+          type: 'tag',
+          name: 'sup',
+          attribs: {
+            'class': 'bar'
```
jenstroeger, line 228: Looks like debug residue?

spec/selection.spec.js
```diff
@@ -218,6 +218,68 @@ describe('Selection', function () {
...
+    })
+
+    describe('subscript:', function () {
+      beforeEach(function () {
+        this.oldSubscriptMarkup = config.subscriptMarkup
+        config.subscriptMarkup = {
+          type: 'tag',
+          name: 'sub',
+          attribs: {
+            'class': 'bar'
```
jenstroeger, line 259: Here too, this looks like debug residue?

jenstroeger:
@TorstenDittmann, any chance you could add two buttons to the [example tooltip](https://github.com/livingdocsIO/editable.js/blob/7dcf471a50461a447a8357d0c8731a1778e8919d/examples/index.js#L47) just for demo purposes of this feature?

TorstenDittmann:
This is actually not necessary, I realized that afterwards. 

You can just use `forceWrap`, subscript and superscript might be too explicit.

jenstroeger:
> This is actually not necessary […]

How so? [`forceWrap()`](https://github.com/livingdocsIO/editable.js/blob/7dcf471a50461a447a8357d0c8731a1778e8919d/src/selection.js#L238-L242) and [`toggle()`](https://github.com/livingdocsIO/editable.js/blob/7dcf471a50461a447a8357d0c8731a1778e8919d/src/selection.js#L116-L120) help to take care of adding/removing elements to/from selected content, whereas your change adds support for [super](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/sup)/[subscript](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/sub) elements. You mean, a user might as well add these elements using `forceWrap()` and `toggle()` herself, no need for your helper helper functions?

TorstenDittmann:
> > This is actually not necessary […]
> 
> How so? [`forceWrap()`](https://github.com/livingdocsIO/editable.js/blob/7dcf471a50461a447a8357d0c8731a1778e8919d/src/selection.js#L238-L242) and [`toggle()`](https://github.com/livingdocsIO/editable.js/blob/7dcf471a50461a447a8357d0c8731a1778e8919d/src/selection.js#L116-L120) help to take care of adding/removing elements to/from selected content, whereas your change adds support for [super](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/sup)/[subscript](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/sub) elements. You mean, a user might as well add these elements using `forceWrap()` and `toggle()` herself, no need for your helper helper functions?

Yeah sorry, thats what I meant 👍

jenstroeger:
Ok, gotcha. I’m a fan of simple helper functions so I thought your contribution is useful. But let’s leave that decision to @peyerluk… 😉

PR #184, opened 2020-08-20
fix keyboard event chaining

TorstenDittmann:
Chained events with spaces like [this](https://github.com/livingdocsIO/editable.js/blob/master/src/dispatcher.js#L208) one were not parsed and did not work.

### DIFF
--- a/src/eventable.js
+++ b/src/eventable.js
@@ -36,8 +36,10 @@ function getEventableModule (notifyContext) {
   let listeners = {}
 
   function addListener (event, listener) {
-    listeners[event] = listeners[event] || []
-    listeners[event].unshift(listener)
+    event.split(' ').forEach(value => {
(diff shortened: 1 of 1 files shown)


src/eventable.js
```diff
@@ -36,8 +36,10 @@ function getEventableModule (notifyContext) {
   let listeners = {}
 
   function addListener (event, listener) {
-    listeners[event] = listeners[event] || []
-    listeners[event].unshift(listener)
+    event.split(' ').forEach(value => {
```
jenstroeger, line 39: To nitpick, I’d rename `event` to `events` because the string can potentially contain more than one event, separated by space.

peyerluk:
Thanks! I cherry-picked your commit, added some tests and fixed it over here:
https://github.com/livingdocsIO/editable.js/pull/191

PR #186, opened 2020-10-22
Fix switch event

peyerluk:
## Motivation

The switch event was broken due to wrong keyboard event handling.

And then there is the issue that the switch event should allow to move the cursor to the previous or next content block in the same way as the browser does when all text is in the same contenteditable.

To do this even after this PR we lack some functionality. We need to store the `x` position when a user first presses either key up or down and then send these coordinates with the switch event once the cursor reaches the boundary of the current contenteditable.

And to finish it up we would need a method to insert the cursor at the beginning or end of a contenteditable at coordinate `x`.

## Changelog

- 🐞 Trigger switch event again
- 🍬 The switch event now triggers when pressing key up in the first line of text and key down at the last line of text. It will also prevent the cursor to jump to the start or end of the line. (Currently we compare y coordinates for this detection. But I'm not 100% sure this is never thrown off by different text sizes or markup like superscript)
- 🔧 Fix React warnings in example page

### DIFF
--- a/examples/events.js
+++ b/examples/events.js
@@ -183,7 +183,7 @@ export default function (editable) {
       if (!isFromFirstExample(elem)) return
       showEvent({
         name: 'split',
-        content: 'Split this block'
+        content: <span>Split this block</span>
       })
     })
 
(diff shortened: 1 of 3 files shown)


jenstroeger:
Thank you @peyerluk for taking care of this!

jenstroeger:
Perhaps related: https://github.com/livingdocsIO/editable.js/pull/186?

PR #187, opened 2020-10-24
Minor fixes to the examples page.

jenstroeger:
This PR contains a bunch of minor fixes to the comment texts on the example page, and I also [updated the use of React’s `PropTypes`](https://reactjs.org/blog/2017/04/07/react-v15.5.0.html#migrating-from-reactproptypes).

@peyerluk, I’ve also added an example that shows how to style the editable but I was unsure whether you’d be interested. Happy to add that here or in a separate PR.

### DIFF
--- a/examples/events.js
+++ b/examples/events.js
@@ -201,7 +201,7 @@ export default function (editable) {
 
     .on('switch', (elem, direction, cursor) => {
       if (!isFromFirstExample(elem)) return
-      const content = direction === 'after'
+      const content = direction === 'down'
         ? 'Set the focus to the following block'
         : 'Set the focus to the previous block'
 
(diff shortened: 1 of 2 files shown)


peyerluk:
Thanks. Looks good :)

The only thing would be to update the commit messages a bit to be in line with [semantic-relase](https://github.com/semantic-release/semantic-release). I can also do this quickly if you prefer.

For example:
Fix a few of the example comments. -> chore(examples): fix a few of the example comments


> I’ve also added an example that shows how to style the editable

I think s separate PR would be best. Then we can merge this PR here.

jenstroeger:
@peyerluk, thank you for the link to the [semantic-release](https://github.com/semantic-release/semantic-release) project, I wasn’t aware of it! Please feel free to squash and modify commit messages as you see fit.

> I think a separate PR would be best.

Will do 👍

peyerluk:
Rebased and merged over here: https://github.com/livingdocsIO/editable.js/pull/189

PR #196, opened 2020-10-29
Hoist helper functions from Selection into Cursor class.

jenstroeger:
This helper function is useful to check whether the cursor is currently inside of a certain element. For example, within links I want to ignore split and linefeed events:
```javascript
      if (cursor.getParentContainer('A')) {
        return;
      }
 ```
@peyerluk, I think you had implemented a similar function in PR https://github.com/livingdocsIO/editable.js/pull/99 regarding links and not wanting to wrap a link in a link. This PR might support yours?

### DIFF
--- a/src/cursor.js
+++ b/src/cursor.js
@@ -37,6 +37,35 @@ export default class Cursor {
     this.isCursor = true
   }
 
+  // Get all tags that affect the current selection. Optionally pass a
+  // method to filter the returned elements.
+  //
+  // @param {Function filter(node)} [Optional] Method to filter the returned
+  //   DOM Nodes.
+  // @return {Array of DOM Nodes}
+  getTags (filterFunc) {
+    return content.getTags(this.host, this.range, filterFunc)
+  }
+
+  // Get the names of all tags that affect the current selection. Optionally
+  // pass a method to filter the returned elements.
+  //
+  // @param {Function filter(node)} [Optional] Method to filter the DOM
+  //   Nodes whose names are returned.
+  // @return {Array<String> of tag names}
+  getTagNames (filterFunc) {
+    return content.getTagNames(this.getTags(filterFunc))
+  }
+
+  // Get all tags of the specified type that affect the current selection.
+  //
+  // @method getTagsByName
+  // @param {String} tagName. E.g. 'a' to get all links.
+  // @return {Array of DOM Nodes}
+  getTagsByName (tagName) {
+    return content.getTagsByName(this.host, this.range, tagName)
+  }
+
   isAtEnd () {
     return parser.isEndOfHost(
       this.host,
--- a/src/selection.js
+++ b/src/selection.js
@@ -249,35 +249,6 @@ export default class Selection extends Cursor {
     this.setSelection()
   }
 
-  // Get all tags that affect the current selection. Optionally pass a
-  // method to filter the returned elements.
-  //
-  // @param {Function filter(node)} [Optional] Method to filter the returned
-  //   DOM Nodes.
-  // @return {Array of DOM Nodes}
-  getTags (filterFunc) {
-    return content.getTags(this.host, this.range, filterFunc)
-  }
-
-  // Get the names of all tags that affect the current selection. Optionally
-  // pass a method to filter the returned elements.
-  //
-  // @param {Function filter(node)} [Optional] Method to filter the DOM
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/cursor.js [hunk 1]
```diff
@@ -289,6 +289,21 @@ export default class Cursor {
     return content.adoptElement(node, this.win.document)
   }
 
+  // Check whether the cursor's closest parent element node matches
+  // the given tagName (and optional className), and return that
+  // parent if it does or null if not.
+  getParentContainer (tagName, className) {
```
peyerluk, line 295: This method only checks the direct parent. Usually in editable.js we collect all ancestors up to the host so you now inside which elements e.g. a cursor is currently in. Only checking the direct parent is not enough in some cases.

We already have `getTags()` and `getTagsByName()` which return all elements up to the host which affect a range.

`getTags()` and `getTagsByName()` are currently only available in the selection. But I think we could simply move them up to the cursor (The `Selection` class extends `Cursor`).

What do you think about this approach?

jenstroeger: @peyerluk, that would also work if it’s guaranteed that they both return an ordered list of matching ancestors. 

Does the performance difference between the two (fetching the closest matching ancestor vs. building a list of all matching ancestors) matter to you?

peyerluk: The performance difference is negligible. 

The order will be reliable for a cursor. In case of the selection there are many cases of tags which surround the selection are contained within it or only start or end in the selection.

jenstroeger: Works for me then 👍🏼

Do you want to make the change, or would you like to close this PR in exchange for a new PR, or should I repurpose this PR?

peyerluk: If you would like it sooner rather than later you can go ahead :) Otherwise I can do this the next time I'll do some coding in editable.js.

jenstroeger: @peyerluk, I hoisted 

https://github.com/livingdocsIO/editable.js/blob/59962fe995b2d7c3c0fdb4015ca9f70d0460dfca/src/selection.js#L244-L261

into [cursor.js](https://github.com/livingdocsIO/editable.js/blob/master/src/cursor.js) and that seems to have worked. With that change, I can now rewrite the new `getParentContainer()` function like so:
```javascript
  // Check whether the cursor's closest parent element node matches
  // the given tagName (and optional className), and return that
  // parent if it does or null if not.
  getParentContainer (tagName, className) {
    const tags = this.getTags()

    let node = tags.shift()
    if (node.nodeType !== Node.TEXT_NODE) return null

    if (tags.length === 0) return null

    node = tags.shift()
    if (node.nodeName !== tagName) return null
    if (className && !node.classList.contains(className)) return null

    return node
  }
```
Should probably be renamed to `getParentTag()` to fit in with other names in this class.

However, I have a feeling that you may not be too enthusiastic about this function because it may be too specific? Either way, happy to

 - either keep and expand this PR and this function with our discussed changes; or
 - close this PR and submit a new one that contains only hoisting the two functions.

jenstroeger: Addendum: when hoisting `getTags()` and `getTagsByName()` from `Selection` into `Cursor` I think I should change my other PR https://github.com/livingdocsIO/editable.js/pull/198/files accordingly and hoist `getTagNames()` into `Cursor` as well 🤔

jenstroeger: Addressed all of the above, still testing some more though.

jenstroeger:
@peyerluk, I rebased on yesterday’s master.

src/cursor.js
peyerluk: I like the hoisting of `getTags()` etc. to Cursor. But I think  the `getParentContainer()` method is too specific to your use case and should not be part of editable.js. This code can easily live in client apps once we expose `getTags()` on the `Cursor` class.

jenstroeger: Fairnuff. Want me to adjust this PR accordingly?

jenstroeger:
I removed the `getParentContainer()` function, and this PR now contains only the hoisted functions in a single commit.

jenstroeger:
@peyerluk, @marcbachmann, when hoisting these three methods from Selection to Cursor I forgot to adjust the comments, e.g. “Get all tags that affect the current selection.” I think that should be “Get all tags that affect the current cursor or selection.”. Do you care, then I’ll add another PR?

PR #198, opened 2020-10-29
Add getTagNames() helper function to selection.

jenstroeger:
I needed this helper function, and perhaps you do, too 🤓

### DIFF
--- a/src/selection.js
+++ b/src/selection.js
@@ -264,7 +264,7 @@ export default class Selection extends Cursor {
   //
   // @param {Function filter(node)} [Optional] Method to filter the DOM
   //   Nodes whose names are returned.
-  // @return {Array of DOM Node names}
+  // @return {Array<String> of tag names}
   getTagNames (filterFunc) {
     return content.getTagNames(this.getTags(filterFunc))
   }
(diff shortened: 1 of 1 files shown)


src/selection.js
```diff
@@ -251,6 +251,16 @@ export default class Selection extends Cursor {
     return content.getTags(this.host, this.range, filterFunc)
   }
 
+  // Get the names of all tags that affect the current selection. Optionally
+  // pass a method to filter the returned elements.
+  //
+  // @param {Function filter(node)} [Optional] Method to filter the DOM
+  //   Nodes whose names are returned.
+  // @return {Array of DOM Node names}
```
peyerluk, line 259: This would be an array of strings right?
e.g. `// @return {Array<String>}`

In general the comments are currently pseudo JSDoc but I would like to slowly move them to proper JSDoc with Typescript syntax where possible as I think that provides the most value nowadays. I tell you this just as info. We'll have to go through the code someday and upgrade all the comments.

jenstroeger: Thank you, darn that copy & paste 😉 See commit https://github.com/livingdocsIO/editable.js/pull/198/commits/cd981e2ae3c5d09a5698c7692a72577675b04eaf.

jenstroeger:
@peyerluk, I rebased the branch on top of the latest commit https://github.com/livingdocsIO/editable.js/commit/bf3bf5c0ba8127bb26f31bde8102e835585b195e.

PR #199, opened 2020-10-29
Add new highlighting change event

meinradandermatt:
# Motivation
The spellcheck is automatically update the highlighting after a change in the editable. This happens asynchronously and can lead to change which we will miss.

# Changelog
🍬fire `spellcheckUpdated` event after spellchecker has updated highlighting

### DIFF
--- a/src/highlighting.js
+++ b/src/highlighting.js
@@ -135,7 +135,6 @@ export default class Highlighting {
 
       this.safeHighlightMatches(editableHost, matchCollection.matches)
     })
-
   }
 
   // Calls highlightMatches internally but ensures
@@ -148,6 +147,9 @@ export default class Highlighting {
       })
     } else {
       this.highlightMatches(editableHost, matches)
+    }
+    if (this.editable.dispatcher) {
+      this.editable.dispatcher.notify('spellcheckUpdated', editableHost)
     }
   }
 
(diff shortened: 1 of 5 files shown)


peyerluk:
Should we add the new event to the docs (README and `examples.index`)?

peyerluk:
I think in this case the name highlightingChange might be misleading. Currently this is more a `highlightingUpdated` or even a `spellcheckUpdated` event as we dont do any change detection and I think it does not apply to other highlights.

meinradandermatt:
> I think in this case the name highlightingChange might be misleading. Currently this is more a `highlightingUpdated` or even a `spellcheckUpdated` event as we dont do any change detection and I think it does not apply to other highlights.

That's a good point. I will rename it to spellcheckUpdated.

src/highlighting.js
```diff
@@ -98,13 +99,13 @@ export default class Highlighting {
   // Manage Highlights
   // -----------------
 
-  editableHasChanged (editableHost, throttle) {
+  editableHasChanged (editableHost, throttle, raiseEvents) {
```
peyerluk, line 102: Why do you pass `raiseEvents` here?

Don't we want to trigger the spellcheck event every the spellcheck highlight has updated?

meinradandermatt: I didn't trigger it on the update of the spellcheck out of the framework. But this is not necessary anymore. So it makes more sense to always trigger it. I will remove it.

peyerluk (approved):

PR #200, opened 2020-10-31
Remove jQuery -  Part 1

TorstenDittmann:
In this PR I replaced jQuery in following files:

- src/block.js 
- src/clipboard.js
- src/core.js 
- src/create-default-behavior.js
- src/cursor.js

and created a helper at `src/util/dom.js`

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -138,7 +138,7 @@ const Editable = module.exports = class Editable {
   disable (target) {
     const document = this.win.document
 
-    target = target ? [target] : Array.from(document.querySelectorAll(`.${config.editableClass}`))
+    target = target ? [target] : [...document.querySelectorAll(`.${config.editableClass}`)]
     target.forEach(element => block.disable(element))
 
     return this
@@ -156,7 +156,7 @@ const Editable = module.exports = class Editable {
     const document = this.win.document
     const shouldSpellcheck = this.config.browserSpellcheck
 
-    target = target || Array.from(document.querySelectorAll(`.${config.editableDisabledClass}`))
+    target = target || [...document.querySelectorAll(`.${config.editableDisabledClass}`)]
     target.forEach(element => {
       block.init(element, {normalize, shouldSpellcheck})
       this.dispatcher.notify('init', element)
@@ -176,7 +176,7 @@ const Editable = module.exports = class Editable {
   suspend (target) {
     const document = this.win.document
 
-    target = target ? [target] : Array.from(document.querySelectorAll(`.${config.editableClass}`))
+    target = target ? [target] : [...document.querySelectorAll(`.${config.editableClass}`)]
     target.forEach(element => element.removeAttribute('contenteditable'))
 
     this.dispatcher.suspend()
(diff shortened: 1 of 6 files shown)


src/util/dom.js
```diff
@@ -0,0 +1,37 @@
...
+ * @param {HTMLElement | Array | String} target
+ * @param {Document} document
+ */
+export const domArray = (target, document) => {
+  let targets = []
+  switch (true) {
+    case isString(target):
+      targets = Array.from(document.querySelectorAll(target))
+      break
+    case target instanceof Array:
```
marcbachmann, line 13: Can we use `Array.isArray(target)` here?
That makes the function more compatible for Array instances of other documents, e.g. an iframe.

src/core.js
```diff
@@ -138,7 +138,7 @@ const Editable = module.exports = class Editable {
   disable (target) {
     const document = this.win.document
 
-    target = target ? [target] : Array.from(document.querySelectorAll(`.${config.editableClass}`))
+    target = target ? [target] : [...document.querySelectorAll(`.${config.editableClass}`)]
```
marcbachmann, line 141: We could also replace the `target.forEach` call with a `for (const element of target)` loop. That way this conversion won't be necessary at all.

src/util/dom.js
TorstenDittmann: Today I learned: `Array.isArray(target)` exists!

Didn't know that before 😄 Gonna adapt my code.

marcbachmann (approved):
Perfect. thank you

src/core.js
```diff
@@ -138,19 +150,19 @@ const Editable = module.exports = class Editable {
...
-
+  enable (target, normalize) {
+    const document = this.win.document
     const shouldSpellcheck = this.config.browserSpellcheck
-    $elem.each((i, el) => {
-      block.init(el, {normalize, shouldSpellcheck})
-      this.dispatcher.notify('init', el)
-    })
+
+    target = target ? [target] : document.querySelectorAll(`.${config.editableDisabledClass}`)
```
peyerluk, line 161: We could use a small helper here which would also keep this working with jquery elements.
```js
function toElementArray (node) {
  return node.jquery ? node.toArray() : [node]
}
```

TorstenDittmann: Sounds reasonable! Will do 👍

TorstenDittmann:
fyi: still working on this, but I am close to my exams right now - so my time is quite limited. Gonna make this jQuery compatible and merge it 👍

peyerluk:
There is no rush. Editable.js is our patient workhorse in the background :)

Good Luck on your exams!

marcbachmann:
oops. that close wasn't on purpose. I deleted the base branch as it didn't have any commits.
image

I can't reopen it here

marcbachmann:
After recreating the branch, the reopen works again

marcbachmann:
I've opened #220, so we can land this

marcbachmann:
@TorstenDittmann are you fine if we continue on there?

PR #202, opened 2020-11-05
Extract specific highlight ranges

meinradandermatt:
# Motivation
At the moment all highlight ranges are returned in the `getHighlightPositions()` function. Now it is possible to get only a specific type of highlights. 

Changelog
🍬add type param to `getHighlightPositions` function to get only highlightRanges from a specific type.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -379,7 +379,7 @@ const Editable = module.exports = class Editable {
   }
 
   /**
-   * Extracts positions of all DOMNodes that match `[data-word-id]`.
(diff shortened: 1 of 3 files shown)


src/core.js
```diff
@@ -388,12 +388,14 @@ const Editable = module.exports = class Editable {
    * ```
    *
    * @param  {Object} options
-   * @param  {DOMNode} options.editableHos
+   * @param  {DOMNode} options.editableHost
+   * @param  {String} options.type
```
peyerluk, line 392: is `type` optional?

peyerluk (approved):

PR #204, opened 2020-12-03
Upgrade node chromium and eslint

marcbachmann:
I just upgraded node and chromium.
And saw that eslint was quite outdated. I also ran `eslint --fix` and fixed some shadowed variables.


.droneci.yml
```diff
@@ -7,17 +7,17 @@ clone:
...
-  image: marcbachmann/eslint:5.10.1
+  image: marcbachmann/eslint:7.14.0
   environment:
     GH_TOKEN:
       from_secret: GH_TOKEN
 
 - name: npm-install
-  image: node:8-alpine
-  commands: [npm install]
+  image: cypress/browsers:node14.15.0-chrome86-ff82
```
marcbachmann, line 16: @peyerluk that chromium is working properly and has root permissions.

.eslintrc.json
```diff
@@ -1,21 +1,199 @@
 {
-  "extends": "standard",
+  "parserOptions": {
+    // Node 14 already supports ES2020, Node 12 only supports ES2019
+    // https://node.green/#ES2020
+    "ecmaVersion": 2020,
```
marcbachmann, line 5: This should be save as we still run webpack with the ES2015 config.

meinradandermatt: Should we force the people to Node 14? For me it's fine but can you please add this to the readme in the development section.

meinradandermatt (approved):
Nice, can you just add the new requirement for node 14 to the PR and to the readme?

package.json
```diff
@@ -50,7 +45,7 @@
     "webpack-dev-server": "^1.14.1"
   },
   "engines": {
-    "node": ">=0.10.0"
+    "node": ">=12"
```
peyerluk, line 48: Is node 12 still fine with the eslint config you added?

peyerluk (approved):

.eslintrc.json
marcbachmann: This is just the tooling, not the final build. there's also an .nvmrc file.
And the comment is only about supported features in the eslint parsing, not in the environment itself. Sorry, didn't mention that.

package.json
marcbachmann: yes, https://github.com/livingdocsIO/editable.js/pull/204#discussion_r535994941

marcbachmann: we could also just upgrade to node 14, but then I'd remove the engines declaration in here, as it shouldn't be in the final module build. This library works in the browser, not on node, so it could report false negatives during an install of a project that depends on node 12. e.g. our projects.

peyerluk: I think node 12 is fine for the moment.

PR #211, opened 2021-05-14
feat(selection): add removeChars() method

peyerluk:
# Changelog

* 🎁 Add method to `selection`: `selection.removeChars(['«', '»', '"'])`

### DIFF
--- a/spec/selection.spec.js
+++ b/spec/selection.spec.js
@@ -41,6 +41,19 @@ describe('Selection', function () {
     })
   })
 
+  describe('removeChars()', function () {
+
+    it('removes multiple characters', function () {
+      this.div = $('<div>«Foo "bar" foo»</div>')[0]
+      const range = rangy.createRange()
+      range.selectNodeContents(this.div)
+      this.selection = new Selection(this.div, range)
+
+      this.selection.removeChars(['«', '»', '"'])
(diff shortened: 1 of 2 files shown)


spec/selection.spec.js
```diff
@@ -41,6 +41,19 @@ describe('Selection', function () {
     })
   })
 
+  describe('removeChars()', function () {
+
+    it('removes multiple characters', function () {
```
meinradandermatt, line 46: it would be nice to have a test that the characters are not removed in the tag `<a href=""></a>`

src/selection.js
```diff
@@ -182,6 +182,15 @@ export default class Selection extends Cursor {
     this.setSelection()
   }
 
+  removeChars (chars = []) {
+    for (let i = 0; i < chars.length; i++) {
+      const char = chars[i]
+      this.range = content.deleteCharacter(this.host, this.range, char)
+
```
meinradandermatt, line 189: remove empty line

meinradandermatt (approved):

spec/selection.spec.js
peyerluk: I didn't add one as the method works on the DOM so these characters don't really exist from that perspective. So I think it's fine for now. We don't currently have methods that work on serialised HTML in the `cursor` or `selection` classes.

PR #212, opened 2021-05-14
replace quotation marks

meinradandermatt:
# Motivation
On paste the quotation marks in a text should be replaced with quotation marks set in the config. 

# Changelog
- 🎁 Add quotation mark replace on paste
added config in `pastedHtmlRules`
```
replaceQuotes: {
      quotes: ['“', '”'],
      singleQuotes: ['‘', '’'],
      apostrophe: '’'
    }
```

### DIFF
--- a/src/clipboard.js
+++ b/src/clipboard.js
@@ -172,7 +172,7 @@ export function cleanWhitespace (str) {
 }
 
 export function replaceAllQuotes (str) {
-  if (replaceQuotes.quotes) {
+  if (replaceQuotes.quotes || replaceQuotes.singleQuotes || replaceQuotes.apostrophe) {
     return quotes.replaceAllQuotes(str, replaceQuotes)
   }
 
--- a/src/clipboard.js
+++ b/src/clipboard.js
@@ -4,11 +4,21 @@ import config from './config'
 import * as string from './util/string'
 import * as nodeType from './node-type'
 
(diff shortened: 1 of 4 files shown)


src/config.js
```diff
@@ -75,6 +75,12 @@ export default {
     // and content (text content and child elements) will get removed.
     blacklistedElements: ['style', 'script'],
 
-    keepInternalRelativeLinks: false
+    keepInternalRelativeLinks: false,
+
+    replaceQuotes: {
```
peyerluk, line 80: I think we should not introduce a default config.

peyerluk (approved):

src/clipboard.js
```diff
@@ -168,3 +170,11 @@ export function cleanWhitespace (str) {
       : ' '
     ))
 }
+
+export function replaceAllQuotes (str) {
+  if (replaceQuotes.quotes) {
```
peyerluk, line 175: Can you alsoadd `if (replaceQuotes &&)` here so we can deal with a completely undefined config?

meinradandermatt: it's always an empty object https://github.com/livingdocsIO/editable.js/pull/212/files#diff-e4b7db0db2bb9f545c59492bc99ef442cd32bc5402335021236b8f159dd3cd06R22

peyerluk: perfect!

PR #213, opened 2021-06-15
Provide cursor offset search and set methods

gabrielhase:
Related:
- https://github.com/livingdocsIO/livingdocs-framework/pull/573
- https://github.com/livingdocsIO/livingdocs-editor/pull/4496

## Motivation

In order to support maintaining the cursor position when moving up and down with the arrow keys in between editable containers (e.g. paragraphs) we implement methods to search a character offset given a coordinate set as well as a method to create a cursor at a given character offset.

```js
const {wasFound, offset, distance} = this.editable.findClosestCursorOffset({editable, element, origCoordinates})
if (wasFound) this.editable.createCursorAtCharacterOffset({element, offset})
```

### 🎁  `findClosestCursorOffset({element, origCoordinates, requiredOnFirstLine, requiredOnLastLine})`

Takes an element containing text nodes and a coordinate box and uses the `left` coordinate to position a character offset in `element` that most closely matches the left coordinate value. Returns an object with `wasFound` to indicate if a character offset could be found and `offset` to indicate the actual offset.
Internally, this uses a binary search to set the cursor at a position, calculating its bounding box, comparing the cursors left value to the given left value and iterating towards the result.

If the `requiredOnFirstLine` or `requiredOnLastLine` flags are given it will also use the `top` coordinate to make sure the character offset is only found on the right line. This is useful when navigating the cursor with the arrow key down from a paragraph to the next (requires the cursor to be on the first line).

### 🎁  `createCursorAtCharacterOffset ({element, offset})`

Creates a cursor in the given element at the given character offset. Returns the created cursor.
Calls `setVisibleSelection` internally.

### 🐞 Fire `switch` event more reliable

In certain situations the switch event was not fired. This depended on how the browser would set the the range of the current cursor. If the range of a cursor was in an element node and not a text node it would not fire. This affected emtpy editable elements the most but occurred also frequently in other situations.

### Binary Cursor Search

Add a helper method to perform a binary cursor search given a setup of methods. See the code file for more documentation.

### Element utils

Add the following util methods to perform operations on elements:
- `textNodesUnder`, returns all text nodes under a given DOM node
- `getTotalCharCount`, count all text node characters under a given DOM node
- `getTextNodeAndRelativeOffset`, takes an absolute character offset and a DOM node and loops through all nested text nodes to find the text node as well as the relative offset in that text node to which the absolute offset translates to

### DIFF
--- a/spec/api.spec.js
+++ b/spec/api.spec.js
@@ -153,7 +153,6 @@ Cursor 2:                                   |
           element: $div[0],
           origCoordinates: {top: 0, left: 130}
         })
-        console.log('offset', offset)
         expect(wasFound).toEqual(true)
         expect(offset).toEqual(19)
       })
--- a/spec/api.spec.js
+++ b/spec/api.spec.js
@@ -126,11 +126,11 @@ describe('Editable', function () {
 
     describe('findClosestCursorOffset:', function () {
     /*
-Cursor1:                     | (left: 130)
-Comp 1:   Cristiano Ronaldo wurde 2018 mit grossem Tamtam nach Turin geholt.
-Comp 2:   Der Spieler blieb bei fünf Champions-League-Titeln stehen.
-Cursor 2:                    | (offset: 19 chars)
-    */
+     * Cursor1:                     | (left: 130)
+     * Comp 1:   Cristiano Ronaldo wurde 2018 mit grossem Tamtam nach Turin geholt.
+     * Comp 2:   Der Spieler blieb bei fünf Champions-League-Titeln stehen.
+     * Cursor 2:                    | (offset: 19 chars)
+     */
       it('finds the index in a text node', function () {
         $div.html('Der Spieler blieb bei fünf Champions-League-Titeln stehen.')
         const {wasFound, offset} = editable.findClosestCursorOffset({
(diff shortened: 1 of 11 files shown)


spec/api.spec.js
```diff
@@ -123,5 +123,63 @@ describe('Editable', function () {
...
+      })
+
+      /*
+Cursor1:                                                   |
+Comp 1:   Cristiano Ronaldo wurde 2018 mit grossem Tamtam nach Turin geholt.
+Comp 2:   Foo
+Cursor 2: not found
+    */
+      it('returns not found for coordinates that are out of the text area', function () {
+        $div.html('Foo')
```
peyerluk, line 176: Shouldn't it return an offset at the end of the text here? Like `offset: 3`?

src/core.js
```diff
@@ -217,6 +219,27 @@ export default class Editable {
     return new Cursor($host[0], range)
   }
 
+  createRangyRange () {
```
peyerluk, line 222: It would be better not to expose the rangy range. Is it still needed by the framework?

src/core.js
```diff
@@ -446,6 +469,86 @@ export default class Editable {
...
+      data.rightLimit = data.currentOffset
+      data.currentOffset = Math.floor((data.currentOffset - data.leftLimit) / 2)
+    }
+    // move the binary search index in between the current position and the right limit
+    const moveRight = (data) => {
+      data.leftLimit = data.currentOffset
+      data.currentOffset = data.currentOffset + Math.floor((data.rightLimit - data.currentOffset) / 2)
+    }
+    // consider a small bluriness since character never exactly match coordinates
+    const foundChecker = ({distance, bluriness}) => {
```
peyerluk, line 520: I think we should simply return the closest cursor position. There is no need for a `bluriness` factor.

src/dispatcher.js
```diff
@@ -165,17 +166,17 @@ export default class Dispatcher {
 
   dispatchSwitchEvent (event, element, direction) {
     if (event.altKey || event.ctrlKey || event.metaKey || event.shiftKey) return
-
     const cursor = this.selectionWatcher.getSelection()
     if (!cursor || cursor.isSelection) return
 
-    if (direction === 'up' && cursor.isAtFirstLine()) {
+    const totalCharCount = getTotalCharCount(element)
+    if (direction === 'up' && (cursor.isAtFirstLine() || totalCharCount === 0)) {
```
peyerluk, line 173: Do you know why `isAtFirstLine()` does not work here?

gabrielhase: What do you mean by "does not work here"?

src/core.js
gabrielhase: No, I don't need this anymore. It was just super handy to have this while trying out things in the framework. Any reason why you don't want people to use rangy?

spec/api.spec.js
gabrielhase: IMO it's better if it tells me that this is not really a fit at all. But you already mentioned that you would return the closest cursor position plus the pixel offset. That is fine too, then I can maintain the behavior in the framework. Important IMO:
- if you switch from a main area paragraph say in the middle to a sidebar title I don't want it to try and match a position nor to put the cursor at the end, but putting  the cursor at the beginning

PR #216, opened 2021-07-02
chore: Update jQuery peer dependency to v3.5.1.

jenstroeger:
I think this is a typo, so fixing the jQuery peer dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
   "license": "MIT",
   "main": "lib/core.js",
   "peerDependencies": {
-    "jquery": "^3.4.1"
+    "jquery": "^3.5.1"
   },
   "repository": {
     "type": "git",
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,7 +57,7 @@
         "node": ">=12"
       },
       "peerDependencies": {
-        "jquery": "^3.4.1"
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -62,7 +62,7 @@
   "license": "MIT",
   "main": "lib/core.js",
   "peerDependencies": {
-    "jquery": "^3.4.1"
+    "jquery": "^3.5.1"
```
marcbachmann, line 65: is there any reason not to upgrade to 3.6.0 right away?

jenstroeger: The only reason for me personally: I’m using [Bootstrap 4.5](https://getbootstrap.com/docs/4.5/) which uses jQuery 3.5.1 ([docs](https://getbootstrap.com/docs/4.5/getting-started/introduction/#bundle)) and therefore Editable has worked just fine in that context.

I haven’t tried to upgrade to [jQuery 3.6.0](https://github.com/jquery/jquery/releases/tag/3.6.0) and I don’t know how that upgrade would work with Bootstrap.

marcbachmann: ah, the reference to the other commit was the missing link

marcbachmann (approved):

PR #217, opened 2021-07-02
chore: Export Editable constructor as default.

jenstroeger:
@peyerluk please correct the commit message if necessary, I was just following your instructions with this one 😉

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -19,6 +19,7 @@ module.exports = {
   output: {
     library: dist ? 'Editable' : undefined,
     libraryTarget: 'umd',
+    libraryExport: 'default',
     path: __dirname,
     filename: '[name].js'
   },


webpack.config.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
   output: {
     library: dist ? 'Editable' : undefined,
     libraryTarget: 'umd',
+    libraryExport: 'default',
```
marcbachmann, line 22: webpack suggests to use the new object based declaration.
Maybe we can use that:
```js
output: {
  library: {
    name: dist ? 'Editable' : undefined, // not sure why we have the condition here
    type: 'umd',
    export: 'default'
  }
}
```

I guess with the above config, imports need to be changed from
```
import * as Editable from 'upfront-editable'
```
to
```
import Editable from 'upfront-editable'
```

In our repos we build from source, [which already exports `Editable` in `src/core.js` as `default`](https://github.com/livingdocsIO/editable.js/blob/master/src/core.js#L38) , so this is safe to change on our side.
But any consumer using `dist/editable.js` might be affected (probably not many). Maybe a major version change would be good.

What bundler do you use where you ran into issues? Do you directly import dist/editable.js?

marcbachmann:
There's an exports related change here: https://github.com/livingdocsIO/editable.js/pull/210

webpack.config.js
jenstroeger: > What bundler do you use where you ran into issues? Do you directly import dist/editable.js?

I copy the `dist/editable.js` into the static assets of my webapp and then include it directly in the html:
```html
<script src="https://code.jquery.com/jquery-3.5.1.min.js" integrity="sha256-9/aliU8dGd2tb6OSsuzixeV4y/faTqgFtohetphbbj0=" crossorigin="anonymous"></script>
<script src="https://stackpath.bootstrapcdn.com/bootstrap/4.5.2/js/bootstrap.bundle.min.js" integrity="sha384-LtrjvnR4Twt/qOuYxE721u19sVFLVSA4hf/rRt6PrZTmiPltdZcI7q7PXQBYTKyf" crossorigin="anonymous"></script>
<script src="https://webapp.server/static/js/editable.js"></script>
<script src="https://webapp.server/static/js/webapp.js"></script>  <!-- More jQuery-based webapp code. -->
```

marcbachmann (approved):

PR #219, opened 2021-07-04
Run npm audit on CI

### LINKED ISSUE #218: Add `npm audit` to CI?
I just ran an
```
> npm audit
...
21 vulnerabilities (5 low, 8 moderate, 3 high, 5 critical)
...
```
and I wonder if it would make sense to add auditing to the [.droneci.yml](https://github.com/livingdocsIO/editable.js/blob/master/.droneci.yml) file? I’m not too familiar with [Drone](https://docs.drone.io/) and the plan you’re using, though.

marcbachmann:
### Changelog
- Move .droneci.yml to .drone.yml (I haven't removed the old file yet to simplify the migration)
- Add `npm audit` to the drone pipeline (fixes #218)
- Enable regular builds for PRs


package.json
```diff
@@ -45,7 +43,7 @@
     "url-loader": "^4.1.1",
     "webpack": "^5.35.0",
     "webpack-cli": "^4.6.0",
-    "webpack-dev-server": "^3.11.2"
+    "webpack-dev-server": "^4.0.0-beta.3"
```
marcbachmann, line 46: This bumps chokidar to a version without security issues

package.json
```diff
@@ -13,7 +13,6 @@
     "@babel/plugin-transform-runtime": "^7.13.15",
     "@babel/preset-env": "^7.13.15",
     "@babel/preset-react": "^7.13.13",
-    "babel-istanbul-loader": "^0.1.0",
```
marcbachmann, line 16: This module didn't get any updates since 4 years 🙈

ajwild (commented):
I'm happy with the changes, and everything worked as expected when testing, however I'd consider building package-lock.json with npm 6 for this in case it causes any compatibility issues.

jenstroeger:
> I'd consider building package-lock.json with npm 6 for this in case it causes any compatibility issues.

I used npm6 and received the following message
```
npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!
```
and a _huge_ diff of the lock file; updating to npm7 removed that message and left the lock file clean.

ajwild (approved):
@jenstroeger this is what we've experienced with a few repositories. I've made a mistake here though - When I saw thousands of lines had changed in package-lock.json I thought it was a version change, but I already update to lockfile v2 [here](https://github.com/livingdocsIO/editable.js/pull/208)! I remember facing a few issues which is why I was being cautious, and I thought we had rolled the change back but that is not the case. As long as all contributors now use npm v7 we should be fine.

PR #220, opened 2021-07-04
Remove jQuery - part 1

marcbachmann:
This PR is a continuation of #200
I'd like to land that partial updates as soon as possible, so we make at least some progress.

### DIFF
--- a/src/clipboard.js
+++ b/src/clipboard.js
@@ -1,5 +1,3 @@
-import $ from 'jquery'
-
 import config from './config'
 import * as string from './util/string'
 import * as nodeType from './node-type'
@@ -43,28 +41,30 @@ export function paste (element, cursor, callback) {
   // After that grab the pasted content, filter it and restore the focus.
   setTimeout(() => {
     const blocks = parseContent(pasteHolder)
-    $(pasteHolder).remove()
+    pasteHolder.parentNode.removeChild(pasteHolder)
     element.removeAttribute(config.pastingAttribute)
     cursor.restore()
     callback(blocks, cursor)
   }, 0)
 }
 
+/**
+ * @param { Document } document
+ */
 export function injectPasteholder (document) {
-  const pasteHolder = $('<div>')
-    .attr('contenteditable', true)
-    .css({
-      position: 'fixed',
-      right: '5px',
-      top: '50%',
-      width: '1px',
-      height: '1px',
-      overflow: 'hidden',
-      outline: 'none'
-    })
+  const pasteHolder = document.createElement('div')
+  pasteHolder.setAttribute('contenteditable', true)
+  pasteHolder.style.position = 'fixed'
+  pasteHolder.style.right = '5px'
+  pasteHolder.style.top = '50%'
+  pasteHolder.style.width = '1px'
+  pasteHolder.style.height = '1px'
+  pasteHolder.style.overflow = 'hidden'
+  pasteHolder.style.outline = 'none'
 
-  $(document.body).append(pasteHolder)
-  return pasteHolder.get(0)
+  document.body.appendChild(pasteHolder)
+
+  return pasteHolder
 }
 
 /**
--- a/src/clipboard.js
+++ b/src/clipboard.js
@@ -41,7 +41,7 @@ export function paste (element, cursor, callback) {
   // After that grab the pasted content, filter it and restore the focus.
   setTimeout(() => {
     const blocks = parseContent(pasteHolder)
-    pasteHolder.parentNode.removeChild(pasteHolder)
+    pasteHolder.remove()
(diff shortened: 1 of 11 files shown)


src/util/dom.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * @param {HTMLElement | Array | String} target
+ * @param {Document} document
+ */
+export const domArray = (target, document) => {
+  if (typeof target === 'string') return Array.from(document.querySelectorAll(target))
+  if (target instanceof HTMLElement) return [target]
+  if (target.jquery) return target.toArray()
+  if (Array.isArray(target)) return target
+  return []
```
marcbachmann, line 10: Maybe we should remove that default array fallback

src/util/dom.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * @param {HTMLElement | Array | String} target
+ * @param {Document} document
+ */
+export const domArray = (target, document) => {
+  if (typeof target === 'string') return Array.from(document.querySelectorAll(target))
+  if (target instanceof HTMLElement) return [target]
+  if (target.jquery) return target.toArray()
```
jenstroeger, line 8: You want to keep _some_ jQuery support around?

jenstroeger (approved):
This change looks good to me, and I’m curious if there’s any measurable performance difference?

Also, @marcbachmann, just out of curiosity: you used 
```javascript
for (const element of targets) { ... }
```
rather than
```javascript
targets.forEach(element => ...)
```
Any particular reason? (More [here](https://stackoverflow.com/questions/25994909/difference-between-foreach-and-for-loop-in-javascript).)

src/util/dom.js
jenstroeger: I think that’s ok, but would probably document it.

src/util/dom.js
marcbachmann: At the moment it's not quite obvious which methods are private and which ones public.
As we don't want to introduce any issues during the migration I'd keep the compatibility with jQuery.
Later we could do a breaking change and just  throw an error here to get a proper stack trace in case somebody passes an unsupported parameter.

marcbachmann:
> This change looks good to me, and I’m curious if there’s any measurable performance difference?

as most of the operations are scoped to a single contenteditable block, I doubt that there's a big improvement.

Did you experience some issues before?
I've only found some bugs already present on master when I tested it on the demo page. 

> Also, @marcbachmann, just out of curiosity: you used
> 
> ```js
> for (const element of targets) { ... }
> ```
> 
> rather than
> 
> ```js
> targets.forEach(element => ...)
> ```
> 
> Any particular reason? (More [here](https://stackoverflow.com/questions/25994909/difference-between-foreach-and-for-loop-in-javascript).)

I use the for loops mainly to improve the readability, consistency and also compatibility of array-like constructs. Sometimes to improve the performance by replacing map & filter constructs, but that's not the case here.

src/util/dom.js
jenstroeger: Makes sense 👍🏼

jenstroeger:
> Did you experience some issues before?

@marcbachmann, no I haven’t. Change looks good!

src/core.js
```diff
@@ -112,9 +114,14 @@ export default class Editable {
...
    */
   remove (target) {
-    const $target = $(target)
-    this.disable($target)
-    $target.removeClass(config.editableDisabledClass)
+    const targets = domArray(target, this.win.document)
+
+    this.disable(targets)
+
+    for (const element of target) {
```
peyerluk, line 121: Shouldn't this be `... of targets`?

src/core.js
```diff
@@ -96,7 +96,9 @@ export default class Editable {
    * @chainable
    */
   add (target) {
-    this.enable($(target))
+    const targets = domArray(target, this.win.document)
```
peyerluk, line 99: `this.enable(target)` would also work here as `domArray()` is called within `enable` again.

src/clipboard.js
```diff
@@ -43,28 +41,30 @@ export function paste (element, cursor, callback) {
   // After that grab the pasted content, filter it and restore the focus.
   setTimeout(() => {
     const blocks = parseContent(pasteHolder)
-    $(pasteHolder).remove()
+    pasteHolder.parentNode.removeChild(pasteHolder)
```
marcbachmann, line 44:
```suggestion
    pasteHolder.remove()
```

peyerluk (approved):

src/util/dom.js
marcbachmann: I've made it more compatible and just replaced it with `Array.from(target)`.
That way it's compatible with `NodeList` and `jQuery` instances out of the box.

PR #222, opened 2021-07-07
docs: Fix typos in comments.

jenstroeger:


jenstroeger:
Also, this:

https://github.com/livingdocsIO/editable.js/blob/616c47241b4df307ee46ab27eecd5466b6cd12b7/src/parser.js#L189-L200

should be “lastChild” (maybe “lastestChild” which isn’t really a word) but I think it’s definitely _not_ [latest](https://dict.leo.org/german-english/latest) child 😉

marcbachmann (approved):

marcbachmann:
> should be “lastChild” (maybe “lastestChild” which isn’t really a word) but I think it’s definitely not latest child 😉

Maybe it's not named `lastChild` because it returns the same element if there are no children.
So I'd either rename that to `lastElement` or `lastChild`. @peyerluk what do you think?

peyerluk:
For me `lastChild` would be ok. We can also name it `lastChildOrSelf` alternatively.

jenstroeger:
Thanks @peyerluk, I’ll submit a separate PR for that.

PR #223, opened 2021-07-08
feat(selection): Add helper functions to select tags.

jenstroeger:
Split the original `content.getTags()` function into two, to make its components available separately. Then make these two new functions available to the Cursor (and, thus, the Selection).

### DIFF
--- a/src/cursor.js
+++ b/src/cursor.js
@@ -65,6 +65,16 @@ export default class Cursor {
     return content.getTagsByName(this.host, this.range, tagName)
   }
 
+  // Get all tags that are completely withing the current selection.
+  getInnerTags (filterFunc) {
+    return content.getInnerTags(this.range, filterFunc)
+  }
+
+  // Get all tags that surround the current selection.
+  getAncestorTags (filterFunc) {
+    return content.getAncestorTags(this.host, this.range, filterFunc)
+  }
+
   isAtEnd () {
     return parser.isEndOfHost(
       this.host,
--- a/src/cursor.js
+++ b/src/cursor.js
@@ -65,7 +65,7 @@ export default class Cursor {
     return content.getTagsByName(this.host, this.range, tagName)
   }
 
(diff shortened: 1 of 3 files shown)


src/cursor.js
```diff
@@ -65,6 +65,16 @@ export default class Cursor {
     return content.getTagsByName(this.host, this.range, tagName)
   }
 
+  // Get all tags that are completely withing the current selection.
```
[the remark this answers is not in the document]
jenstroeger: Commit https://github.com/livingdocsIO/editable.js/pull/223/commits/1dfdfd5b536483eed86d70590e433026388a6e95.

peyerluk (approved):

peyerluk (approved):

src/cursor.js
```diff
@@ -65,6 +65,16 @@ export default class Cursor {
     return content.getTagsByName(this.host, this.range, tagName)
   }
 
+  // Get all tags that are completely within the current selection.
+  getInnerTags (filterFunc) {
```
peyerluk, line 69: These new functions should have a test each.

jenstroeger: Makes sense. The tests would go into [cursor.spec.js](https://github.com/livingdocsIO/editable.js/blob/master/spec/cursor.spec.js) file?

peyerluk:
There are two linting errors which should be fixed before we can merge (this is why the CI fails).

src/cursor.js
jenstroeger: Commit https://github.com/livingdocsIO/editable.js/pull/223/commits/057c3158cff780ed4f52b21d6d64c25bdb07a2e1 contains two (very) simple tests.

jenstroeger:
@peyerluk @marcbachmann I don’t know why this PR’s checks are stuck, can you please check? It’d be great to merge.

PR #224, opened 2021-07-08
feat(core): Allow to add/remove a handler to multiple events.

jenstroeger:
Similar to jQuery’s [on()](https://api.jquery.com/on/) and [off()](https://api.jquery.com/off/) function, this change enables users to register a handler function for more than one event.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -433,33 +433,38 @@ export default class Editable {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


peyerluk:
As we don't have an event object where you can see which event was raised I don't want to add the possibility to subscribe to multiple events at once at this moment.

PR #226, opened 2021-07-08
feat(selection): Add isWrappable() helper function.

jenstroeger:


peyerluk:
Out of curiosity: what use case do you need this method for?

jenstroeger:
In the `"selection"` event handler I check whether the given selection can be wrapped at all, and only if it can be wrapped does a little toolbar pop open.

jenstroeger:
Any chance we can merge this one, too, please? 😇

jenstroeger:
@peyerluk @marcbachmann I don’t know why this PR’s checks are stuck, can you please check? It’d be great to merge.

marcbachmann:
It's just the eslint reporting that doesn't work for pull requests as the token doesn't get injected for security reasons.

PR #228, opened 2021-07-08
feat(selection): Remove nodes based on tag name and class.

jenstroeger:
Extend the [selection.removeFormatting()](https://github.com/livingdocsIO/editable.js/blob/77e45729651eda118ec232c0adf603171d7686b0/src/selection.js#L202-L206) function so that it can remove nodes based on tag name _and_ class name; handy if I’d like to remove e.g. `span.foo` only.

### DIFF
--- a/src/content.js
+++ b/src/content.js
@@ -303,19 +303,17 @@ export function removeFormattingElem (host, range, elem) {
   })
 }
 
-export function removeFormatting (host, range, tagName, selector) {
+export function removeFormatting (host, range, selector) {
   return restoreRange(host, range, () => {
-    nuke(host, range, tagName, selector)
+    nuke(host, range, selector)
   })
 }
 
 // Unwrap all tags this range is affected by.
 // Can also affect content outside of the range.
-export function nuke (host, range, tagName, selector) {
+export function nuke (host, range, selector) {
   getTags(host, range).forEach((elem) => {
-    if (elem.nodeName.toUpperCase() !== 'BR' &&
-      (!tagName || (elem.nodeName.toUpperCase() === tagName.toUpperCase() &&
-        (!selector || elem.matches(selector))))) {
+    if (elem.nodeName.toUpperCase() !== 'BR' && (!selector || elem.matches(selector))) {
       unwrap(elem)
     }
   })
--- a/src/content.js
+++ b/src/content.js
@@ -303,19 +303,19 @@ export function removeFormattingElem (host, range, elem) {
   })
 }
 
-export function removeFormatting (host, range, tagName, className) {
+export function removeFormatting (host, range, tagName, selector) {
   return restoreRange(host, range, () => {
-    nuke(host, range, tagName, className)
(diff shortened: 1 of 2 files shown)


src/content.js
```diff
@@ -310,18 +310,19 @@ export function removeFormattingElem (host, range, elem) {
...
 
 // Unwrap all tags this range is affected by.
 // Can also affect content outside of the range.
-export function nuke (host, range, tagName) {
+export function nuke (host, range, tagName, className) {
   getTags(host, range).forEach((elem) => {
     if (elem.nodeName.toUpperCase() !== 'BR' &&
-      (!tagName || elem.nodeName.toUpperCase() === tagName.toUpperCase())) {
+      (!tagName || (elem.nodeName.toUpperCase() === tagName.toUpperCase() &&
+        (!className || elem.classList.contains(className))))) {
```
marcbachmann, line 325: We could use a selector here instead of a fixed class name. That way you can pass `.someclass` and also attributes `[data-someattribute]`.

jenstroeger: Great suggestion, thanks @marcbachmann! Commit https://github.com/livingdocsIO/editable.js/pull/228/commits/1f008690ce35f9304d63910a4577a8cc7cb14f1c.

peyerluk (approved):

src/content.js
```diff
@@ -310,18 +310,19 @@ export function removeFormattingElem (host, range, elem) {
   })
 }
 
-export function removeFormatting (host, range, tagName) {
+export function removeFormatting (host, range, tagName, selector) {
```
peyerluk, line 313: We could shorten this to: `export function removeFormatting (host, range, selector) {`

This would still be compatible with the old version.

jenstroeger: Good point, thanks @peyerluk! Commit https://github.com/livingdocsIO/editable.js/pull/228/commits/56491acacdaecd2537eea8197ccafd809ca7f35a.

src/content.js
```diff
@@ -310,18 +310,17 @@ export function removeFormattingElem (host, range, elem) {
...
 }
 
 // Unwrap all tags this range is affected by.
 // Can also affect content outside of the range.
-export function nuke (host, range, tagName) {
+export function nuke (host, range, selector) {
   getTags(host, range).forEach((elem) => {
-    if (elem.nodeName.toUpperCase() !== 'BR' &&
-      (!tagName || elem.nodeName.toUpperCase() === tagName.toUpperCase())) {
+    if (elem.nodeName.toUpperCase() !== 'BR' && (!selector || elem.matches(selector))) {
```
jenstroeger, line 323: An alternative implementation of that function would be:
```javascript
export function nuke (host, range, selector) {
  if (!selector) selector = "br"
  getTags(host, range).forEach((elem) => {
    if (elem.matches(selector)) unwrap(elem)
  })
}
```

marcbachmann (approved):

PR #231, opened 2021-07-20
fix(events): Check event.target.closest() exists before calling

ajwild:
In Firefox 89 I'm getting the following error when selecting or deselecting components:
```
Uncaught TypeError: evt.target.closest is not a function
    getEditableBlockByEvent webpack://@livingdocs/editor/../editable.js/lib/dispatcher.js?:271
```

This doesn't make much sense, as `closest()` has been supported in Firefox [since version 35](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#browser_compatibility), but this little check is enough to handle the issue.

### DIFF
--- a/src/dispatcher.js
+++ b/src/dispatcher.js
@@ -4,6 +4,7 @@ import eventable from './eventable'
 import SelectionWatcher from './selection-watcher'
 import config from './config'
 import Keyboard from './keyboard'
+import {closest} from './util/dom'
 
 // This will be set to true once we detect the input event is working.
 // Input event description on MDN:
@@ -30,6 +31,7 @@ export default class Dispatcher {
     this.keyboard = new Keyboard(this.selectionWatcher)
     this.activeListeners = []
     this.setup()
+    this.getEditableBlockByEvent = (evt) => closest(evt.target, editable.editableSelector)
   }
 
   setupDocumentListener (event, func, capture = false) {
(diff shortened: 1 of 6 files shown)


src/dispatcher.js
```diff
@@ -82,7 +82,8 @@ export default class Dispatcher {
   }
 
   getEditableBlockByEvent (evt) {
-    return evt.target.closest(this.editableSelector)
+    // For some unknown reason `closest` does not always exist in Firefox 89 ¯\_(ツ)_/¯
+    return evt.target.closest ? evt.target.closest(this.editableSelector) : undefined
```
marcbachmann, line 86: I suppose this is caused by text nodes beimg the event target.
```js
return (evt.target.closest ? evt.target : evt.target.parentNode).closest(this.editableSelector)
```

ajwild: It seems that if you click outside of the iframe then the target is the `HTMLDocument` and there is no `closest()` or `parentNode`.

A few observations with "initial" interaction blocker:
- On click the function is not called. This is probably due to `preventDefault()` being called elsewhere.
- On blur within the iframe the function is not called.
- On blur outside the iframe the function is called with the iframe body as the target (and the component is not deselected).
- On blur to another component the function is not called (but is called for the focus on the other component).

PR #233, opened 2021-07-28
Introduce boundary selection event

gabrielhase:
## Motivation

In order to support selections over multiple editable directives (e.g. over multiple paragraphs) we introduce the new event type `selectToBoundary` that is fired when:
- a selection is at the start of an editable directive (going up)
- a selection is at the end of an editable directive (going down)
- or a selection is at both ends (start and end)

Related Framework PR: https://github.com/livingdocsIO/livingdocs-framework/pull/582

### 🎁 Introduce `selectToBoundary` event

The `selectToBoundary` event fires when a text selection is dragged to a boundary of an editable: either at the start, at the end or at both ends (if e.g. all text of a paragraph is selected).
The event passes the following params: `(element, event, position)`
- element is the DOM node on which the event is called
- event is the native selection event
- position is either `start`, `end` or `both` depending on which boundary the event fires

### 🍬  support selectionchange for Firefox

Firefox (Gecko) supports the selectionchange event since 2017. Time to enable it in editable.js for this browser :)

### 🍬  allow to pass highlight type to highlight

Previously the highlight code always set the type to `comment` thus also resulting in a CSS class `highlight-comment`. We now allow to pass the type which is useful if you want to define other kinds of highlights (and according CSS classes, e.g. `selection` -> `highlight-selection`).
```
const type = 'custom'
this.editable.highlight({editableHost, text, highlightId, textRange, raiseEvents, type})
// -> will result in a highlight with css class `highlight-custom`
```

### 🐞  normalize DOM node after removing highlight

After removing a highlight (span tag) we need to normalize the parent DOM node otherwise the DOM node will still be split into 2 children (at least in Chrome).

### DIFF
--- a/src/feature-detection.js
+++ b/src/feature-detection.js
@@ -16,13 +16,11 @@ const browserEngine = parser.getEngineName()
 const webKit = browserEngine === 'WebKit'
 
 /**
- * Check selectionchange event (currently supported in IE, Chrome and Safari)
- *
- * To handle selectionchange in firefox see CKEditor selection object
- * https://github.com/ckeditor/ckeditor-dev/blob/master/core/selection.js#L388
+ * Check selectionchange event (currently supported in IE, Chrome, Firefox and Safari)
+ * Firefox supports it since version 52 (2017) so pretty sure this is fine.
  */
 // not exactly feature detection... is it?
-export const selectionchange = !(browserEngine === 'Gecko' || browserName === 'Opera')
+export const selectionchange = !(browserName === 'Opera')
 
 // See Keyboard.prototype.preventContenteditableBug for more information.
 export const contenteditableSpanBug = !!webKit
--- a/src/feature-detection.js
+++ b/src/feature-detection.js
@@ -11,16 +11,30 @@ import browser from 'bowser'
 export const contenteditable = typeof document.documentElement.contentEditable !== 'undefined'
 
 const parser = browser.getParser(window.navigator.userAgent)
-const browserName = parser.getBrowser()
 const browserEngine = parser.getEngineName()
 const webKit = browserEngine === 'WebKit'
 
 /**
- * Check selectionchange event (currently supported in IE, Chrome, Firefox and Safari)
- * Firefox supports it since version 52 (2017) so pretty sure this is fine.
+ * Check selectionchange event (supported in IE, Chrome, Firefox and Safari)
+ * Firefox supports it since version 52 (2017).
+ * Opera has no support as of 2021.
  */
-// not exactly feature detection... is it?
-export const selectionchange = !(browserName === 'Opera')
+const hasNativeSelectionchangeSupport = (document) => {
+  const doc = document
+  const osc = doc.onselectionchange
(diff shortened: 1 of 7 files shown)


src/feature-detection.js
```diff
@@ -16,13 +16,11 @@ const browserEngine = parser.getEngineName()
...
- * Check selectionchange event (currently supported in IE, Chrome and Safari)
- *
- * To handle selectionchange in firefox see CKEditor selection object
- * https://github.com/ckeditor/ckeditor-dev/blob/master/core/selection.js#L388
+ * Check selectionchange event (currently supported in IE, Chrome, Firefox and Safari)
+ * Firefox supports it since version 52 (2017) so pretty sure this is fine.
  */
 // not exactly feature detection... is it?
-export const selectionchange = !(browserEngine === 'Gecko' || browserName === 'Opera')
+export const selectionchange = !(browserName === 'Opera')
```
marcbachmann, line 23: We could replace this with a feature detection:
https://github.com/2is10/selectionchange-polyfill/blob/master/selectionchange.js#L42-L54

```
function hasNativeSupport (doc = window.document) {
  var osc = doc.onselectionchange;
  if (osc !== undefined) {
    try {
      doc.onselectionchange = 0
      return doc.onselectionchange === null
    } catch (e) {
    } finally {
      doc.onselectionchange = osc
    }
  }
  return false
}
```

gabrielhase: I would vote for not supporting Opera anymore. What do you think @peyerluk ?

marcbachmann: i thought newer opera versions  are based on webkit. 
from wikipedia;
> In 2013, Opera switched from the Presto engine to Chromium.

marcbachmann: I don't see any reason to support browsers that can't keep up with the spec. I'dimplement the feature detection here but drop support for legacy browsers soon. 

@TorstenDittmann what browser versions do you need to support?

TorstenDittmann: Nothing too specific, technically I need to support Chrome and Safari to support Electron/Capacitor.

The rest I don't really mind to be fair +:1:

src/highlight-support.js
```diff
@@ -69,6 +69,8 @@ const highlightSupport = {
     const elems = editableHost.querySelectorAll(`[data-word-id="${highlightId}"]`)
     for (const elem of elems) {
       content.unwrap(elem)
+      // in Chrome browsers the unwrap method leaves the host node split into 2 (lastChild !== firstChild)
```
marcbachmann, line 72: We should always assume that there are empty text nodes.

gabrielhase: sure, but the normalize call doesn't hurt in that case. or are you refering to something else?

marcbachmann: I just womdered whether there's some hidden logic that depends on it. If there's no need I wouldn't apply it as it's a bit inconsistent if we do it only for that event.

peyerluk (approved):

PR #234, opened 2021-07-30
Use null escape instead of br tags for empty elements

marcbachmann:
Before:
| Without Focus | With Focus  |
|-----------------|:-----------:|
| image      | image    |

After:
| Without Focus | With Focus  |
|-----------------|:-----------:|
| image  | image |

The cursor also shows up correctly in Chrome & Safari & Firefox.

### DIFF
--- a/src/create-default-behavior.js
+++ b/src/create-default-behavior.js
@@ -24,13 +24,11 @@ export default function createDefaultBehavior (editable) {
 
   return {
     focus (element) {
-      // Add a <br> element if the editable is empty to force it to have height
-      // E.g. Firefox does not render empty block elements and most browsers do
-      // not render  empty inline elements.
+      // Add an empty text node if the editable is empty to force it to have height
+      // E.g. Firefox does not render empty block elements
+      //   and most browsers do not render empty inline elements.
       if (!parser.isVoid(element)) return
-      const br = document.createElement('br')
-      br.setAttribute('data-editable', 'remove')
-      element.appendChild(br)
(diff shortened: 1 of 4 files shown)


src/create-default-behavior.js
```diff
@@ -24,13 +24,11 @@ export default function createDefaultBehavior (editable) {
...
-      // E.g. Firefox does not render empty block elements and most browsers do
-      // not render  empty inline elements.
+      // Add an empty text node if the editable is empty to force it to have height
+      // E.g. Firefox does not render empty block elements
+      //   and most browsers do not render empty inline elements.
       if (!parser.isVoid(element)) return
-      const br = document.createElement('br')
-      br.setAttribute('data-editable', 'remove')
-      element.appendChild(br)
+      element.appendChild(document.createTextNode('\u0000'))
```
jenstroeger, line 31: I didn’t know! I’ve seen this happening before that an empty text node was folded away and best I could come up with was a `&nbsp;` but that too sometimes failed to work.

jenstroeger (approved):
Thanks @marcbachmann, that’s a neat trick with the null character!

In a separate PR would it be possible to expand the README with more detailed information on the test framework?

src/create-default-behavior.js
marcbachmann: Using `\u200B` would result in the same behavior, but we're using that zero width space already for something else: https://github.com/livingdocsIO/editable.js/blame/master/src/content.js#L87

peyerluk (approved):
Nice, nice, nice!

PR #236, opened 2021-07-31
Fix newline behavior

marcbachmann:
### Changelog
- 🐛 Do not remove newlines in between child nodes of editable blocks
- 🐛 Do not require two shift+return when appending a new line at the end of an editable block
- 🍬 Trim newlines at the start and end of editable blocks

Behavior Before:
Type text, shift return, shift return, type text. blur
| Before Blur | After Blur  |
|-----------------|:-----------:|
| image | image  |

After:
Type text, shift return, type text. blur
Results in 1 exactly newline being present.
And only one shift return is necessary to trigger a newline.

Some tests would still be nice

### DIFF
--- a/src/create-default-behavior.js
+++ b/src/create-default-behavior.js
@@ -44,11 +44,29 @@ export default function createDefaultBehavior (editable) {
     },
 
     newline (element, cursor) {
-
+      // When the cursor is at the text end, we'll need to add an empty text node
+      // after the br tag to ensure that the cursor shows up on the next line
       if (cursor.isAtTextEnd()) {
-        const trailingBr = document.createElement('br')
-        trailingBr.setAttribute('data-editable', 'remove')
-        cursor.insertBefore(trailingBr)
+        // We need to wrap the newline, so it can get deleted
+        // together with the null escape character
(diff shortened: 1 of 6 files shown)


src/create-default-behavior.js
```diff
@@ -46,11 +46,29 @@ export default function createDefaultBehavior (editable) {
...
+
+        // The null escape character gets wrapped in another span,
+        // so it gets removed automatically.
+        // contenteditable=false prevents a focus of the span
+        // and therefore also prevents content from getting written into it.
+        // If this attribute is defined on the parent wrapper element,
+        // the cursor positioning behaves weird with deletion of the newline
+        const spacer = document.createElement('span')
+        spacer.setAttribute('data-editable', 'remove')
+        spacer.setAttribute('contenteditable', 'false')
```
peyerluk, line 65: 👏 This is some clever trickery here...

peyerluk (approved):

marcbachmann:
I'm not sure whether that will reintroduce the issue we had with https://github.com/livingdocsIO/editable.js/pull/172
But it's not the case with newer versions of safari/chrome.

PR #239, opened 2021-09-06
Improve newline behavior

marcbachmann:
### Changelog
- 🐛  Use zero width spaces instead of null escape characters as they caused issues with some fonts
- 🐛 Use a more elegant solution to force a line break after a `<br>` tag. Previously backspace deleted a character too much when a newline at the end of a block got deleted.

### DIFF
--- a/spec/dispatcher.spec.js
+++ b/spec/dispatcher.spec.js
@@ -245,31 +245,20 @@ describe('Dispatcher', function () {
         expect(elem.innerHTML).to.equal('<br>\uFEFF')
       })
 
-      it('creates a nested br element with a data-editable="unwrap" attribute', () => {
+      it('appends a zero-width space after the br tag to force a line break', () => {
         typeKeys(elem, 'foobar')
         shiftReturn(elem)
         expect(elem.innerHTML).to.equal(
-          `\uFEFF` +
-          `foobar` +
-          `<span data-editable="unwrap">` +
-            `<br><span data-editable="remove" contenteditable="false">\uFEFF</span>` +
-          `</span>`
+          `\uFEFFfoobar<br>\uFEFF`
         )
       })
 
-      it('removes the nested br element when adding a newline afterwards', () => {
+      it('does not append another zero-width space when one is present already', () => {
         typeKeys(elem, 'foobar')
         shiftReturn(elem)
         shiftReturn(elem)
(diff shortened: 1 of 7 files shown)


jenstroeger:
@marcbachmann, without trying out the branch locally, I have a related question. Last week I noticed that if I have an editable element
```html
<p contenteditable="true" spellcheck="true">bla</p>  <!-- Removed data-* attributes for brevity. -->
```
an I press `Shift` + `Enter` at the end of the line to insert a newline doesn’t have a visual effect but the event is fired and my `<br>` is inserted.

I wanted to play around some more with that behavior before bringing it to your attention. Does this change address that?

marcbachmann:
That got fixed already in #236 (two shift + returns were necessary).
This PR now improves the behavior by appending a zero whitespace character as the custom tag that I've introduced was not working properly with `backspace`.

spec/dispatcher.spec.js
```diff
@@ -242,34 +242,23 @@ describe('Dispatcher', function () {
...
         expect(elem.innerHTML).to.equal(
-          `\u0000` +
-          `foobar` +
-          // this is the nested br element that got unwrapped
-          `<br>` +
-          // This is the new nested newline of the second shift + return
-          `<span data-editable="unwrap">` +
-            `<br><span data-editable="remove" contenteditable="false">\u0000</span>` +
-          `</span>`
+          `\uFEFFfoobar<br><br>\uFEFF`
```
marcbachmann, line 261: I've tried to write a test for the deletion, but it's not really possible to trigger a backspace or `InputEvent` of type `deleteContentBackward` that mutates the content using javascript 😢.
We might need to simulate it using manual operations, which basically means reimplementing the backspace behavior.

jenstroeger:
> That got fixed already in #236 (two shift + returns were necessary).

Hmm… I’ve fetched my fork in August and it included that change. The behavior is still there though. Let me fetch the latest updates and try again.

marcbachmann:
Weird. I've tested it in chrome, safari & firefox. And a quite old version of chromium.

jenstroeger:
I’ll try to build a repro case. If that doesn’t work we can hop on a screen share and I can show you and debug?

marcbachmann:
Yes, definitely

PR #241, opened 2021-09-07
Fix copy-paste of bold text

romankaravia:
## Motivation

When copy-pasting formatted text which contains a `<strong>` tag, the tag would get lost in Chrome/Safari (but not in Firefox).

This is fixed by changing the method of retrieving pasted text from the clipboard.
Before: A hidden element with `contenteditable` attribute is focused on paste, then the clipboard content is extracted from there.
New: Clipboard content is taken from the [ClipboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/ClipboardEvent/clipboardData).

The new way seems to work reliably in all browsers and was also manually tested in Chrome, Firefox, Safari, Edge (all on macOS).

I'm not sure why specifically the `<strong>` tag was broken, maybe it's related to [this chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=546461).

After pasting, editable.js filters tags based on the allowedElements configuration. This PR does not change anything about the filtering, but note that the `<strong>` tag is [allowed by default](https://github.com/livingdocsIO/editable.js/blob/v3.0.1/src/config.js#L46) and the bug was not related to this.

## Screenshots

(the first line is copy-pasted to the second line)

Before fix:
Screenshot 2021-09-07 at 14 22 01

After fix:
Screenshot 2021-09-07 at 14 21 35

## Changelog

- 🐞- Copy-pasting formatted text which contains a `<strong>` tag is now working as expected on all browsers.

### DIFF
--- a/src/dispatcher.js
+++ b/src/dispatcher.js
@@ -128,6 +128,7 @@ export default class Dispatcher {
         const block = this.getEditableBlockByEvent(evt)
         if (!block) return
         const afterPaste = (blocks, cursor) => {
+          evt.preventDefault()
           if (blocks.length) {
             this.notify('paste', block, blocks, cursor)
 
@@ -140,7 +141,8 @@ export default class Dispatcher {
         }
 
         const cursor = this.selectionWatcher.getFreshSelection()
-        clipboard.paste(block, cursor, afterPaste)
+        const clipboardContent = evt.clipboardData.getData('text/html')
+        clipboard.paste(block, cursor, clipboardContent, afterPaste)
       })
       .setupDocumentListener('input', function inputListener (evt) {
         const block = this.getEditableBlockByEvent(evt)
--- a/src/dispatcher.js
+++ b/src/dispatcher.js
@@ -127,22 +127,21 @@ export default class Dispatcher {
       .setupDocumentListener('paste', function pasteListener (evt) {
         const block = this.getEditableBlockByEvent(evt)
         if (!block) return
-        const afterPaste = (blocks, cursor) => {
-          evt.preventDefault()
-          if (blocks.length) {
-            this.notify('paste', block, blocks, cursor)
 
-            // The input event does not fire when we process the content manually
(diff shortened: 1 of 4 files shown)


src/dispatcher.js
```diff
@@ -140,7 +141,8 @@ export default class Dispatcher {
         }
 
         const cursor = this.selectionWatcher.getFreshSelection()
-        clipboard.paste(block, cursor, afterPaste)
+        const clipboardContent = evt.clipboardData.getData('text/html')
```
marcbachmann, line 144: Is `text/html` always present? What happens with paste using `CMD + SHIFT + v`?
In that case is it also present or do we need a fallback to `text/plain`?

src/dispatcher.js
```diff
@@ -140,7 +141,8 @@ export default class Dispatcher {
...
+        clipboard.paste(block, cursor, clipboardContent, afterPaste)
```
marcbachmann, line 145: Should we migrate to a synchronous api already?
It always needs to be synchronous anyways, otherwise `evt.preventDefault()` won't have any effect.

```js
const blocks = clipboard.paste(block, cursor, clipboardContent)
```

src/dispatcher.js
marcbachmann: list of media types: https://www.w3.org/TR/clipboard-apis/#reading-from-clipboard

romankaravia: Yes, very good point, with only `text/html`, nothing happened on Cmd + Shift + v, so I added a fallback for `text/plain`.

src/dispatcher.js
romankaravia: Yes, good idea, see last commit. (`cursor` can be reassigned inside the function, so I believe it also needs to be returned together with blocks)

marcbachmann (approved):
Good after fixing the tests to use `expect(...).to.equal(...)`

PR #245, opened 2021-10-05
Fix cursor move rounding issues

### LINKED ISSUE #244: Cursor event doesn’t always move blocks.
Using Chrome I get the following sequence of events on the example page by navigating the cursor up or down the [contenteditable](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable) paragraph:

chrome

This behavior begins to change when I zoom the page, e.g. I don’t get the “Set the focus to the following block” when I zoom to 110% or 90% — I suspect that zoom messes with the bounding box of the host element when the [curser detects whether its on the last line](https://github.com/livingdocsIO/editable.js/blob/ad5e43330e0f2b32687c86346775766167593cb5/src/cursor.js#L93-L100)?

With Safari (I’m using v14.0.1) and not even using zoom, however, I don’t see _any_ of these events:

safari

Instead, the cursor just jumps to the beginning (when moving “up”) or the end (when moving “down”) of the host element.

marcbachmann:
Partially fixes #244 

### Changelog
- 🐛 Fix block focus events when using a browser zoom. There were rounding issues

Safari still has an issue with newlines at the end of an html element.

This doesn't work:
```html
<p>
  Hello
</p>
<p>
  World
</p>
```

And this is working:
```html
<p>Hello</p>
<p>World</p>
```

### DIFF
--- a/src/cursor.js
+++ b/src/cursor.js
@@ -96,7 +96,7 @@ export default class Cursor {
     hostRange.collapse(false)
     const hostCoords = getRangeBoundingClientRect(hostRange, this.win)
     const cursorCoords = getRangeBoundingClientRect(this.range.nativeRange, this.win)
-    return hostCoords.bottom === cursorCoords.bottom
+    return isCloseTo(hostCoords.bottom, cursorCoords.bottom)
   }
 
   isAtFirstLine () {
@@ -105,7 +105,7 @@ export default class Cursor {
     hostRange.collapse(true)
     const hostCoords = getRangeBoundingClientRect(hostRange, this.win)
     const cursorCoords = getRangeBoundingClientRect(this.range.nativeRange, this.win)
-    return hostCoords.top === cursorCoords.top
+    return isCloseTo(hostCoords.top, cursorCoords.top)
   }
 
   isAtBeginning () {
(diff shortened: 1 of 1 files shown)


src/cursor.js
```diff
@@ -368,3 +368,9 @@ function getRangeBoundingClientRect (range, win) {
   el.remove()
   return coords
 }
+
+function isCloseTo (a, b) {
+  if (a === b) return true
+  if (Math.abs(a - b) <= 2) return true
```
jenstroeger, line 374: @marcbachmann, the `2` here is an arbitrary heuristic you made up, yes?

Also, I’m always torn between readable code (yours) and more compact expressions, e.g.
```javascript
function isCloseTo (a, b) {
    return a === b || Math.abs(a - b) <= 2
}
```

jenstroeger (approved):
Thank you for tending to this so quickly! Looks good to me.

src/cursor.js
marcbachmann: just wanted to comment on that, but got drawn into a meeting.
Yes, it is. We could increase it to something like 5.
Not sure yet how the safari calculations work and why there's a 2px diff.
We're placing an element in the same div, so it should be really the same.

jenstroeger: My intuition says to keep the number small, so I think `2` makes sense for now.

Alternatively, we’d make that a config option that defaults to `2` if we wanted more flexibility. But that’s probably a bit over-engineered.

marcbachmann: regarding the style, I don't see any benefit of the shorter version 😁
correctness is more important than compactness.
and correctness can be reached better with simpler code

jenstroeger: Agreed on all three points 🤓

PR #246, opened 2021-10-05
Remove whitespaces at start and end of blocks during initialization

marcbachmann:
### Motivation
Content blocks were able to start with spaces before a word. This causes issues in safari as it's not possible to create cursors and get the correct position at the end.
Removing them before initialization results in more tidy content and is most likely ok.

`<p contenteditable=true>      Hello World</p>`
will result in
`<p contenteditable=true>Hello World</p>`

but only when `editable.enable(elem, true)` is used.

During extraction we apply the same logic but always.

This is another fix for #244

### Changelog
- 🐛 To prevent safari getClientBoundingRect issues, we'll need to remove whitespaces at the start and end of a content block

### DIFF
--- a/src/content.js
+++ b/src/content.js
@@ -29,6 +29,12 @@ export function tidyHtml (element) {
 export function normalizeTags (element) {
   const fragment = document.createDocumentFragment()
 
+  // Remove line breaks at the beginning of a content block
+  removeWhitespaces(element, 'firstChild')
+
+  // Remove line breaks at the end of a content block
+  removeWhitespaces(element, 'lastChild')
+
   for (const node of element.childNodes) {
     // skip empty tags, so they'll get removed
     if (node.nodeName !== 'BR' && !node.textContent) continue
@@ -136,7 +142,7 @@ export function cloneRangeContents (range) {
   return fragment
 }
 
-function removeWhitespaces (node, type) {
+function removeWhitespaces (node, type, firstCall = true) {
   let elem
   while ((elem = node[type])) {
     if (elem.nodeType === nodeType.textNode) {
@@ -145,10 +151,17 @@ function removeWhitespaces (node, type) {
     } else if (elem.nodeName === 'BR') {
       elem.remove()
     } else {
-      if (elem[type]) removeWhitespaces(elem, type)
+      if (elem[type]) removeWhitespaces(elem, type, false)
       break
     }
   }
+
+  if (!firstCall) return
+  elem = node[type]
+  if (elem?.nodeType !== nodeType.textNode) return
+  // Remove whitespaces at the end or start of a block with content
+  //   e.g. '  Hello world' > 'Hello World'
+  elem.textContent = elem.textContent.replace(type.startsWith('last') ? /\s+$/ : /^\s+/, '')
 }
 
 // Remove elements that were inserted for internal or user interface purposes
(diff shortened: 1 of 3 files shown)


jenstroeger:
Hmm, I feel a little apprehensive about this change because it modifies the content of the host element (e.g. the paragraph) — what if the author intends leading or trailing whitespaces?

peyerluk (approved):

marcbachmann:
We could also just trim spaces and not `&nbsp;`?
Or just the ones at the end, which already fixes the safari issue in the demo.
If it needs a config, I'd leave it out completely. As we don't need it in our setup.

jenstroeger:
So, I played around a little using Chrome and Firefox on Mac. On the [example site](https://livingdocsio.github.io/editable.js/) you have this paragraph:

https://github.com/livingdocsIO/editable.js/blob/e07311419ac00b351b5760da9fbec8144b8fe7a3/examples/index.html#L34-L36

Notice that there are several spaces between the `<p>` and the beginning of the text `Nulla …` which _aren’t_ rendered. Now, on the  hit <kbd>Backspace</kbd> when the cursor is right before the first word a `Change` event fires. However, visually, I had expected a `Merge`. Looking at the DOM itself, notice a singular space:

editable

After removing these leading spaces manually in the DOM, the correct event fires. Alternatively, adding [white-space: pre-wrap](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space#values) to the paragraph exposes the spaces and events work as expected.

I stumbled upon this problem when I had to deal with a bunch of paragraphs, each of which is _intentionally_ indented by two spaces. (See above [comment](https://github.com/livingdocsIO/editable.js/pull/246#issuecomment-935645996).) Here too, these leading spaces were not rendered, but they’re part of the `#text` node in the DOM but don’t show in the editor…

<hr>

Another oddity, but this seems to happen only in Firefox.

Again, in the example page: place the cursor at the end of the second paragraph `...eget nulla. |` (notice the trailing single space). Hit <kbd>Backspace</kbd> and _both_ the space and the period get deleted. Then insert the period `.` again and hit <kbd>→</kbd> and the cursor moves right to the period and another space — where did that come from? You see the even sequence in the image:

editable-ff

meinradandermatt:
@marcbachmann why is this a fix? We changed the behavior for customers. So this is a new feature. We already had 2 issues with this behavior and it changed in a release as we apply the patch versions automatically in the framework.

marcbachmann:
The selection didn't work properly before. See https://github.com/livingdocsIO/editable.js/issues/244
If there are whitespaces, cursor operations aren't possible anymore. Maybe we could only restrict it to certain whitespaces.

jenstroeger:
> Or we need a config for it which is disabling the behavior by default.

Yes please. Removing whitespaces isn’t a good idea for elements with [CSS `white-space: pre`](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space), see my comment [above](https://github.com/livingdocsIO/editable.js/pull/246#issuecomment-944838037).

src/content.js
```diff
@@ -145,10 +151,17 @@ function removeWhitespaces (node, type) {
...
       break
     }
   }
+
+  if (!firstCall) return
+  elem = node[type]
+  if (elem?.nodeType !== nodeType.textNode) return
+  // Remove whitespaces at the end or start of a block with content
+  //   e.g. '  Hello world' > 'Hello World'
+  elem.textContent = elem.textContent.replace(type.startsWith('last') ? /\s+$/ : /^\s+/, '')
```
marcbachmann, line 164: `/ +$/ : /^ +/, ''`

PR #249, opened 2021-11-16
fix isElement function

meinradandermatt:
# Motivation
The isElement function has added a direct dependency to the ownerDocument. This is the reason why for the spellcheck the instanceof will fail. 

# Changelog
🐞 support marker in highlighting with other ownerdocument

### DIFF
--- a/spec/highlighting.spec.js
+++ b/spec/highlighting.spec.js
@@ -111,6 +111,25 @@ describe('Highlighting', function () {
       const matches = textSearch.findMatches(blockText, [text])
       expect(matches).to.equal(undefined)
     })
+
+    it('does not go into an endless loop without a html marker node', function () {
+      const blockText = 'Mehr als 90 Prozent der Fälle in Grossbritannien in den letzten vier Wochen gehen auf die Delta-Variante zurück. Anders als bei vorangegangenen Wellen scheinen sich jedoch die Fallzahlen von den Todesfällen und Hospitalisierungen zu entkoppeln.'
+      const text = ''
+      const textSearch = new WordHighlighter('something', 'text')
+      const matches = textSearch.findMatches(blockText, [text])
+      expect(matches).to.equal(undefined)
+    })
+
+    it('handle the marker with other ownerdocument correctly', function () {
+      const blockText = 'Mehr als 90 Prozent'
+      const text = 'Mehr als 90 Prozent'
+      const ifrm = window.document.createElement('iframe')
+      window.document.body.append(ifrm)
+      const markerNode = createElement('<span class="highlight"></span>', ifrm.contentWindow)
+      const textSearch = new WordHighlighter(markerNode, 'text')
+      const matches = textSearch.findMatches(blockText, [text])
+      expect(matches[0].match).to.equal(text)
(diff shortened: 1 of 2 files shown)


jenstroeger:
> In Chrome the isElement function is not working as the HTMLElement is not an object there.

Can you please elaborate on “is not working”? What’s your expected and actual result?

I ask because as per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement): “The `HTMLElement` interface represents any [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) element” and “[it] inherits properties from its parent, Element.” 

There’s also a difference between [typeof](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof) and [instanceof](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof) in JavaScript. I think the
```javascript
typeof HTMLElement === 'object'
```
will always be false because `HTMLElement` is a constructor (i.e. function) on Chrome, Firefox, and Safari:
```
> typeof HTMLElement
< "function"
```

marcbachmann:
I also think the code should work. Maybe we have an issue in our downstream code that the element isn't from the same window and therefore a different `HTMLElement` instance?

marcbachmann:
I'd rather check `typeof obj && obj instanceof obj.ownerDocument?.defaultView?.HTMLElement`

meinradandermatt:
> I also think the code should work. Maybe we have an issue in our downstream code that the element isn't from the same window and therefore a different `HTMLElement` instance?
The code work, I just added the update from stackoverflow (where the existing code is coming from) to test. Yes, it is from a different window. As we have added with this code a dependency to the window. So we have added with that a breaking change. I will check what we can do with the spellcheck. 

Just the example which is not working.

`const createElement = (html, win = window) => {
  const el = win.document.createElement('div')
  el.innerHTML = html
  return el.firstElementChild
}

var ifrm = window.document.createElement("iframe")
window.document.body.append(ifrm)

const spanElement = createElement('<span>test</span>', ifrm.contentWindow)
spanElement.setAttribute('data-editable', 'ui-unwrap')
spanElement.setAttribute('data-highlight', 'something')


spanElement instanceof HTMLElement`  // --> false

marcbachmann (approved):

spec/highlighting.spec.js
```diff
@@ -111,6 +111,25 @@ describe('Highlighting', function () {
...
+      const text = ''
+      const textSearch = new WordHighlighter('something', 'text')
+      const matches = textSearch.findMatches(blockText, [text])
+      expect(matches).to.equal(undefined)
+    })
+
+    it('handle the marker with other ownerdocument correctly', function () {
+      const blockText = 'Mehr als 90 Prozent'
+      const text = 'Mehr als 90 Prozent'
+      const ifrm = window.document.createElement('iframe')
```
marcbachmann, line 126: `const iframe` would be also ok to use. 😉

meinradandermatt: it's def. too long. 😉

PR #251, opened 2021-12-06
Allow formats to be set over links if the beginning or end is at a boundary

gabrielhase:
Issue: https://github.com/livingdocsIO/livingdocs-planning/issues/4333

## Motivation

A link is removed when a bold tag is created around. It happens only when bold surrounds more than the link.

editablewithlink

The problem is actually more general. Whenever a formatting tag is selected directly at the beginning (leading) or end (trailing) the marker will be set *inside* of the format tag thus destroying the HTML markup.

### :beetle: Check format tag position over range

We use the selection range of the browser to check if the selection is directly leading or trailing a formatting tag. If so, we manually set the marker tag directly before or after the format tag in order to keep the HTML markup intact.

## Testing

Manually tested on:
- Mac OSX Chrome
- Mac OSX Safari
- Mac OSX Firefox
- Windows 11 Edge

### DIFF
--- a/examples/index.css
+++ b/examples/index.css
@@ -28,6 +28,13 @@ body {
 }
 
 /*
+    Make trailing whitespaces visible since they can be quite confusing.
+*/
+[data-editable] {
+    white-space: pre-wrap;
+  }
+
+/*
  * Remove text-shadow in selection highlight: h5bp.com/i
  * These selection declarations have to be separate.
  * Customize the background color to match your design.
--- a/examples/index.html
+++ b/examples/index.html
@@ -28,12 +28,8 @@
         <h2 class="example-title">An editable paragraph</h2>
 
         <div class="paragraph-example example-sheet">
-          <p>
-            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer nec odio. Praesent libero. Sed ante dapibus diam. Sed nisi.
-          </p>
-          <p>
-            Nulla quis sem at nibh elementum imperdiet. Duis sagittis ipsum. Praesent mauris. Fusce nec tellus sed augue semper porta. Mauris massa. Vestibulum lacinia arcu eget nulla.
-          </p>
+          <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer nec odio. Praesent libero. Sed ante dapibus diam. Sed nisi.</p>
+          <p>Nulla quis sem at nibh elementum imperdiet. Duis sagittis ipsum. Praesent mauris. Fusce nec tellus sed augue semper porta. Mauris massa. Vestibulum lacinia arcu eget nulla.</p>
         </div>
 
         <div>
@@ -50,16 +46,12 @@
         <h2 class="example-title">Text Formatting</h2>
 
         <div class="formatting-example example-sheet">
-          <p>
-            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer nec odio. Praesent libero.
-          </p>
+          <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer nec odio. Praesent libero.</p>
         </div>
 
         <div class="code-example">
           <h3>HTML</h3>
-<pre><code class="formatting-code-js language-markup">&lt;p&gt;
-  Lorem ipsum dolor sit amet...
-&lt;/p&gt;</code></pre>
+<pre><code class="formatting-code-js language-markup">&lt;p&gt;Lorem ipsum dolor sit amet...&lt;/p&gt;</code></pre>
 
         </div>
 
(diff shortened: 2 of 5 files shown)


jenstroeger:
@gabrielhase, the issue URL doesn’t work.

This change seems loosely related the the leading/trailing spaces issue in https://github.com/livingdocsIO/editable.js/pull/246#issuecomment-944838037?

gabrielhase:
@jenstroeger it absolutely is. I just hit the same issue and in my code would now need to check if there are trailing whitespaces (btw. there can also be a trailing `\n`). I'll probably check first if I can find reason for the trailing invisible characters before proceeding here. Thanks for the input.

jenstroeger:
@gabrielhase I added 
```css
[data-editable] {
  white-space: pre-wrap;
}
```
to my CSS to show whatever whitespace characters exist inside of an edited element. I’d like to keep the difference between server-side text and client-side editing to a minimum.

src/range-save-restore.js
```diff
@@ -11,17 +11,37 @@ let boundaryMarkerId = 0
...
 const markerTextChar = '\ufeff'
 
+function isChildOf (parent, possibleChild) {
+  for (let i = 0; i < parent.children.length; i++) {
+    if (parent.children[i] === possibleChild) return true
+  }
+  return false
+}
+
+function isChildAtBeginning (parent, possibleChild) {
```
jenstroeger, line 21: Why not `isFirstChild()`?

src/range-save-restore.js
```diff
@@ -11,17 +11,37 @@ let boundaryMarkerId = 0
...
+  const isChild = isChildOf(parent, possibleChild)
+  return isChild && parent.childNodes[0] === possibleChild
+}
+
+function isChildAtEnd (parent, possibleChild) {
```
jenstroeger, line 26: Why not `isLastChild()`?

src/range-save-restore.js
```diff
@@ -11,17 +11,37 @@ let boundaryMarkerId = 0
...
+
 export function insertRangeBoundaryMarker (range, atStart) {
   const container = range.commonAncestorContainer
+  let trailsStart = false
+  let trailsEnd = false
+  if (isChildAtEnd(range.startContainer.parentElement, range.endContainer.parentElement)) {
+    trailsEnd = true
+  } else if (isChildAtBeginning(range.endContainer.parentElement, range.startContainer.parentElement)) {
+    trailsStart = true
+  }
```
jenstroeger, line 39: Maybe less verbose and avoids the `if` statements:
```javascript
const hasLeading = isChildAtBeginning(range.endContainer.parentElement, range.startContainer.parentElement)
const hasTrailing = isChildAtEnd(range.startContainer.parentElement, range.endContainer.parentElement)
```

jenstroeger (commented):
Minor comments.

marcbachmann:
@gabrielhase that PR might be related https://github.com/livingdocsIO/editable.js/pull/246

examples/index.css
```diff
@@ -27,6 +27,13 @@ body {
     line-height: 1.4;
 }
 
+/*
+    Make trailing whitespaces visible since they can be quite confusing.
+*/
+[data-editable] {
+    white-space: pre-wrap;
+  }
```
jenstroeger, line 35: Space before ` }` and I’m very tempted to make this part of PR https://github.com/livingdocsIO/editable.js/pull/246 (which has already been merged 😳).

examples/index.html
```diff
@@ -28,12 +28,8 @@ <h2>friendly contenteditable API</h2>
         <h2 class="example-title">An editable paragraph</h2>
 
         <div class="paragraph-example example-sheet">
-          <p>
-            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer nec odio. Praesent libero. Sed ante dapibus diam. Sed nisi.
-          </p>
-          <p>
-            Nulla quis sem at nibh elementum imperdiet. Duis sagittis ipsum. Praesent mauris. Fusce nec tellus sed augue semper porta. Mauris massa. Vestibulum lacinia arcu eget nulla.
-          </p>
```
jenstroeger, line 36: With PR https://github.com/livingdocsIO/editable.js/pull/246 merged, this shouldn’t be a problem anymore?

src/range-save-restore.js
```diff
@@ -11,17 +11,35 @@ let boundaryMarkerId = 0
...
 const markerTextChar = '\ufeff'
 
+function isChildOf (parent, possibleChild) {
+  for (let i = 0; i < parent.children.length; i++) {
+    if (parent.children[i] === possibleChild) return true
+  }
+  return false
+}
+
+function startContainerIsChild (range) {
```
jenstroeger, line 21: “first” instead of “start”?

src/range-save-restore.js
```diff
@@ -11,17 +11,35 @@ let boundaryMarkerId = 0
...
+  const parent = range.endContainer.parentElement
+  const possibleChild = range.startContainer.parentElement
+  return isChildOf(parent, possibleChild)
+}
+
+function endContainerIsChild (range) {
```
jenstroeger, line 27: “last” instead of “end”?

jenstroeger (commented):
Curious what @marcbachmann and @peyerluk think of this. Minor comments below.

src/range-save-restore.js
gabrielhase: I'd rather have this consistent with the naming of the native ranges https://developer.mozilla.org/de/docs/Web/API/Range
they use `start` and `end`

meinradandermatt (approved):

PR #252, opened 2021-12-15
introduce trim config for formats

gabrielhase:
Issue: https://github.com/livingdocsIO/livingdocs-planning/issues/4538

## Motivation

So far editable.js made no special handling for leading or trailing whitespaces in a selection. While in general, it is smart to not make too many assumptions for some formats leading or trailing whitespaces make very little sense. The initial case of this issue was that links leave an empty `<a>` tag with a whitespace if the "visible" part of the link is deleted. While it would be possible to use the `del` and `backspace` event to clean up such cases after every delete IMO this is the wrong way to see this.
There are formats for which leading and trailing whitespaces make no sense. For this reason we introduce a `trim` config to the standard formats that if  set to `true` will not consider leading or trailing whitespaces in the first place. In the example of a link the following selection ` foo bar ` would trim and produce `<a href="..">foo bar</a>`.
There are also formats for which the case is not so clear. For example an underline is visible and at least the case could be made that we want to underline leading and trailing whitespaces. For our custom tags such as "Viertelgefiert", etc. it is explicitly wanted to wrap whitespaces.

### 🎁  Introduce `trim` config on standard formats

We introduce a config `trim` on the standard formats. On the link this would look as follows:
```
linkMarkup: {
    type: 'tag',
    name: 'a',
    attribs: {},
    trim: true
  }
```

If `trim` is set to true then the selection is trimmed before the format is applied. For example setting a link on ` foo bar ` would produce `<a href="..">foo bar</a>`.
The default setting is to set `trim` to `true` for `bold`, `italic`, `link` and false for `underline`. We don't introduce this property for custom tags as this seems too dangerous.

## Testing

Manually tested on:

- Mac OSX Chrome
- Mac OSX Safari
- Mac OSX Firefox
- Windows 11 Edge

### DIFF
--- a/spec/selection.spec.js
+++ b/spec/selection.spec.js
@@ -359,6 +359,55 @@ describe('Selection', function () {
 
   })
 
+  describe('triming:', function () {
+    beforeEach(function () {
+      this.wordWithWhitespace = createElement('<div> foobar </div>')
+      const range = rangy.createRange()
+      range.selectNodeContents(this.wordWithWhitespace.firstChild)
+      this.selection = new Selection(this.wordWithWhitespace, range)
+
+      this.oldLinkMarkup = config.italicMarkup
+      config.linkMarkup = {
+        type: 'tag',
+        name: 'a',
+        attribs: {
+          'class': 'foo bar'
+        },
+        trim: true
+      }
+
+      this.oldUnderlineMarkup = config.underlineMarkup
+      config.underlineMarkup = {
+        type: 'tag',
+        name: 'u',
+        attribs: {
+          'class': 'bar'
+        },
+        trim: false
+      }
+    })
+
+    afterEach(function () {
+      config.linkMarkup = this.oldLinkMarkup
+      config.underlineMarkup = this.oldUnderlineMarkup
+    })
+
+    it('trims whitespaces from range when linking', function () {
+      this.selection.link('https://livingdocs.io')
+      console.log('tag', this.wordWithWhitespace)
+      const linkTags = this.selection.getTagsByName('a')
+      const html = getHtml(linkTags[0])
+      expect(html).to.equal('<a class="foo bar" href="https://livingdocs.io">foobar</a>')
+    })
+
+    it('does not trim whitespaces from range when underlining', function () {
+      this.selection.makeUnderline()
+      const underlineTags = this.selection.getTagsByName('u')
+      const html = getHtml(underlineTags[0])
+      expect(html).to.equal('<u class="bar"> foobar </u>')
+    })
+  })
+
   describe('inherits form Cursor', function () {
 
     it('has isAtEnd() method from Cursor in its protoype chain', function () {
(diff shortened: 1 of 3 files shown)


benib:
For reference:
`\S` is short for `\f\n\r\t\v\u00A0\u2028\u2029`
`String.trim` works on https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#prod-WhiteSpace and https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#prod-LineTerminator

gabrielhase:
@benib I absolutely agree, we should use the same logic for start and end. I used `\S` and `\s` now (they are complementary). A quick search in our customers installations found the following special whitespaces: U+2002, U+2005 and U+2006. `\s` covers: `\f\n\r\t\v​\u00A0\u1680​\u180e\u2000​\u2001\u2002​\u2003\u2004​ \u2005\u2006​\u2007\u2008​\u2009\u200a​\u2028\u2029​\u2028\u2029​ \u202f\u205f​\u3000` and `\S` the inverse of this, so our current cases (and more) are well covered.

spec/selection.spec.js
```diff
@@ -294,6 +294,66 @@ describe('Selection', function () {
 
   })
 
+  describe('triming:', function () {
+    beforeEach(function () {
```
meinradandermatt, line 298: add a case when only a whitespace is selected

src/selection.js
```diff
@@ -149,31 +161,37 @@ export default class Selection extends Cursor {
 
   makeBold () {
```
meinradandermatt, line 162: I think we should extend the toggleCustom as well

spec/selection.spec.js
gabrielhase: Good idea. What would you expect to happen if a selection only consists of whitespace (one or more)? IMO we should then just leave it alone and not trim anything. Fine for you?

src/selection.js
gabrielhase: We can do this. How would you configure it?
From the top of my head it could look like this:
```
customElements: [{
  handle: 'foo',
  ...,
  trim: true | false
}]
```

Since for custom elements this is a project config entry though the change would also be in the server, editor and framework. Is this what you imagine here?

meinradandermatt: I see we don't really have any config in the editablejs for the custom element. So either we would add it with a tag and attributes as an indicator or we would add an option to the toggleCustom function to trim it. I think that would be the best way then we could add the config to the project config on our side.

I also just saw that we can't overwrite the trim config out of our editor.

meinradandermatt (approved):

src/selection.js
```diff
@@ -87,9 +87,21 @@ export default class Selection extends Cursor {
...
       }
     }
+    if (config.linkMarkup.trim) this.trimRange()
+
     this.forceWrap(link)
   }
 
+  // trims whitespaces on the left and right of a selection, i.e. what you want in case of links
+  trimRange () {
+    const textToLink = this.range.toString()
```
jenstroeger, line 97: A bit late but: why `Link` here, shouldn’t the variable be called `textToTrim`?

gabrielhase: thanks you're absolutely right: https://github.com/livingdocsIO/editable.js/pull/259

PR #266, opened 2022-03-09
fix: replace &nbsp; with normal space on 'paste' event

Muhameti-centralsoft:
Relations:

- Issue: https://github.com/livingdocsIO/livingdocs-planning/issues/4587

# Motivation

If on an editable you add a space and then paste some text, the space added previously will be converted on a `non breaking space`.

# Changelog

- 🐞 Replace `non breaking spaces` with normal spaces when pasting on an `editable`.


meinradandermatt (changes requested):
looks good, can you add a test for it?

PR #268, opened 2022-03-29
fix(comment index): index is found with search in added case

JSchenk8:
Relations:
  - Issue: If a word was repeated in a component all copies of that word were highlighted. Furthermore, if a comment was added and then text was added to the component before the highlighted comment the highlight moved

  - Related PR's: [Editor PR](https://github.com/livingdocsIO/livingdocs-editor/pull/5153)

# Motivation

NZZ reported the case where the same word in a component was highlighted repeatedly when an article was re-oponed after highlighting and resolving. When investigating this, it turned out there were several bugs related to the saving of the start index of the comments. This PR combined with the above Editor PR aims to fix this. 

So the use cases where this bug happens is as follows:

- A user comments on a word and resolves the comment
- A user reopens the article or refreshes the page, and reopens the comment
- If there are more than one version of that word in the component then all of them are highlighted

- Alternatively, a comment is highlighted and resolved, the component is altered to have more text added, and then the comment is resolved
- This will change the position of the highlihgt. 
- . 

Multiple word bug:
https://user-images.githubusercontent.com/76001984/160699422-bda46ae7-7e1f-42cf-9d7a-58e23ac88f0b.mov

Behaviour now:

https://user-images.githubusercontent.com/76001984/160699614-acff6ec6-a637-4036-8cc8-4e962d3beab4.mov

The limit to this fix is, if the word is repeated many times in close succession and one is commented, and text is added before the word, and then the comment is resolved, the page refreshed or the article re-entered (a rather rare edge case), then the comment will be highlighted on a different copy of the word.

# Changelog

- 🐞 Multiples of a highlighted word are not highlighted when comment is added and resolved after refresh

### DIFF
--- a/src/highlight-support.js
+++ b/src/highlight-support.js
@@ -12,7 +12,6 @@ const highlightSupport = {
 
   highlightText (editableHost, text, highlightId, type) {
     if (this.hasHighlight(editableHost, highlightId)) return
-
     const blockText = highlightText.extractText(editableHost)
 
     const marker = `<span class="highlight-${type}"></span>`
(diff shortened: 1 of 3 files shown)


src/highlight-support.js
```diff
@@ -28,12 +27,29 @@ const highlightSupport = {
...
+
+  // This function was changed to track matches when text is added to the start of a component, but multiple white spaces break it in a strict sense
+  // The function works in the editor and in browsers, but tests with multiple white spaces will fail.
+  // Browsers change the white spaces to &nbsp and the function works, and the tests in highlight.spec.js have been updated to represent this.
+
+  highlightRange (editableHost, text, highlightId, startIndex, endIndex, dispatcher, type = 'comment') {
     if (this.hasHighlight(editableHost, highlightId)) {
       this.removeHighlight(editableHost, highlightId)
     }
+    const blockText = highlightText.extractText(editableHost)
```
meinradandermatt, line 39: You can move this find match and move the index behavior to a own function so it is easier to understand.

meinradandermatt: You can also add a note to this function that you will find only matches which are bigger than the given index.

meinradandermatt (changes requested):

src/highlight-support.js
JSchenk8: Nice one, I've updated it, let me know what you think

meinradandermatt (approved):

peyerluk:
@JSchenk8 @meinradandermatt 

I plan to revert the code that fixes the index of a highlight as this only works in some use cases but not in others and makes the highlight code hard to understand.

I'm talking mostly about this code here which I removed in my PR:
```js
// ...
const blockText = highlightText.extractText(editableHost)
const matchesArray = textSearch.findMatches(blockText, [text])
const {actualStartIndex, actualEndIndex} = this.getIndex(matchesArray, startIndex, endIndex)
```

For the use case of reopening comments so that the highlight is applied again we should solve this in the Editor. 

We can have a look at code in the editor to search for the commented text when reopening a comment. Also when archiving a comment we should update the context, so that when a comment is reopened we have the correct text to search for.

peyerluk:
This is the PR that removes the behavior:
https://github.com/livingdocsIO/editable.js/pull/394

PR #271, opened 2022-06-12
fix(cutandpaste): Cursor not moved

JSchenk8:
Relations:

Backport: release-2022-05 and release 2022-03
Issue: When anything was pasted into an editable with a space at the end it was pasted at the end of the component
Documentation: -
Related PR's: https://github.com/livingdocsIO/editable.js/pull/266

Motivation

NZZ noticed that when an editor cut from the end of a component and pasted it, sometimes it was pasted back to where it was cut from. This was an unexpected side effect of Muhamet's fix to remove the non breaking space at the end of editables. 

https://user-images.githubusercontent.com/76001984/173227449-178d9247-792c-4b38-b986-51349b933227.mov

Furthermore, Sternwald reported that for some users when using shift + enter properties were added to <br> tags. This is fixed with the updated check in `utils/string.js`. 

The test is a bit superfluous, but checks that the NBSP is removed and that the copied text still exists in the event.

Changelog
🐞 Pasting text into components with a space at the end works correctly and no properties are added to new lines

Cheat Sheet: https://github.com/livingdocsIO/team/blob/master/Coding/Collaboration/Pull-Request-Cheatsheet.md

### DIFF
--- a/spec/dispatcher.spec.js
+++ b/spec/dispatcher.spec.js
@@ -383,7 +383,9 @@ describe('Dispatcher', function () {
       it(`replaces the last '&nbsp' with ' ' if text ends with a single '&nbsp'`, (done) => {
         on('paste', (block, blocks) => {
           expect(block.innerHTML).to
-            .equal('some text that ends with a single a non breaking space ')
+            .equal('some text that ends with a single a non breaking space ') // space has been removed
+          expect(blocks.length).to.equal(1)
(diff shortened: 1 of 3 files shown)


ajwild (changes requested):
This solves the paste bug, but introduces a bug where the cursor moves to the start of the directive on/after paste.

ajwild (approved):

PR #273, opened 2022-08-08
fix: pasting text over selected text which contains markups

Muhameti-centralsoft:
# Relations
  - Issue: https://tracking.nzzmg.ch/browse/WCMS-6398

# Description

There are two issues(videos below) related with pasting text over a selected text which contains markups like `<strong>` or `<em>`

https://user-images.githubusercontent.com/80449613/183460347-e33b442c-5426-472b-97bc-cef7f1b41c73.mov

https://user-images.githubusercontent.com/80449613/183459341-3d2662f3-4f37-487b-aca0-6ab11f48f473.mov

# Changelog

- 🐞 When we paste text while some other text is selected, check If the selected text is exactly surrounded by markup, and if yes delete that markup as well.
- 🐞 When we paste text while some other text is selected, check if there are markups whose text is completely within the current selection, and if yes delete that markup as well.

### DIFF
--- a/src/selection.js
+++ b/src/selection.js
@@ -246,16 +246,30 @@ export default class Selection extends Cursor {
   }
 
   // Delete the farest ancestor that is an exact selection
+  //
+  // @return Selection instance
   deleteExactSurroundingMarkups () {
-    const markupTags = this.getAncestorTags(
+    const ancestorMarkupTags = this.getAncestorTags(
       (elem) => ['STRONG', 'EM'].includes(elem.nodeName)
     ).reverse()
-    for (const markupTag of markupTags) {
-      if (this.isExactSelection(markupTag)) {
-        markupTag.remove()
+    for (const ancestorMarkupTag of ancestorMarkupTags) {
+      if (this.isExactSelection(ancestorMarkupTag)) {
+        ancestorMarkupTag.remove()
         break
       }
     }
+    return new Selection(this.host, this.range)
+  }
+
+  // Delete all the markups whose text is completely within the current selection.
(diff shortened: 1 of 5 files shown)


src/selection.js
```diff
@@ -245,8 +245,35 @@ export default class Selection extends Cursor {
...
+    }
+    return new Selection(this.host, this.range)
+  }
+
+  // Delete all the markups whose text is completely within the current selection.
+  //
+  // @return Selection instance
+  deleteContainedMarkupTags () {
+    const containedMarkupTags = this.getContainedTags(
+      (elem) => ['STRONG', 'EM'].includes(elem.nodeName)
```
meinradandermatt, line 269: Why do you have here only an array of strong and em?

src/selection.js
```diff
@@ -245,8 +245,35 @@ export default class Selection extends Cursor {
     this.setSelection()
   }
 
-  // Delete the contents inside the range. After that the selection will be a
-  // cursor.
+  // Delete the farest ancestor that is an exact selection
```
meinradandermatt, line 248: Can you add some tests for the new functions?

meinradandermatt (changes requested):

src/selection.js
Muhameti-centralsoft: I thought that those two are the only formattings that we can add on the text of an editable.

Muhameti-centralsoft: I removed that filter so now it deletes all kind of tags

meinradandermatt (approved):

PR #282, opened 2022-12-12
Prevent cursor loss on tab switch

peyerluk:
Fixes: https://github.com/livingdocsIO/livingdocs-planning/issues/4773

## Changelog

- 🐞 Prevent cursor loss on tab switch
  
## Background

When switching tabs on coming back a blur event is raised on the editable.  After that the browser sets the cursor again. In our blur event we call `cleanInternals()` wich resets `innerHTML` which causes the browser to loose the cursor position and it sets the cursor to the beginning.
To prevent this we can use `document.activeElement` to check if the editable which receives the blur event is still the active element itself. If so we do an early abort and do not call `cleanInternals()`.


benib (approved):
👍

PR #394, opened 2023-08-11
Remove rangy

marcbachmann:


src/content.js
```diff
@@ -433,3 +435,72 @@ export function nukeTag (host, range, tagName) {
...
+}
+
+function isNodeFullyContained (node, range) {
+  const nodeRange = document.createRange()
+  nodeRange.selectNodeContents(node)
+  return range.compareBoundaryPoints(Range.START_TO_START, nodeRange) <= 0 &&
+         range.compareBoundaryPoints(Range.END_TO_END, nodeRange) >= 0
+}
+
+function canSurroundContents (range) {
```
marcbachmann, line 483: This does not work properly

peyerluk:
> Verify paste cursor move issue (might be related to highlighting)

What is this issue? 
I quickly tested pasting in the editor and for just pasting text everything seems normal (but that's just the simplest test case).

src/content.js
peyerluk: Can you describe this in a bit more detail so I could test this. Or was it unrelated to this PR anyway (still happy to test if you have use cases in mind)?

peyerluk: Ah, I saw that or a very similar bug too!

I'll give it a shot and try to fix it in this PR.

peyerluk:
The happy case is only happening for spellchecking use cases as there we check on every change if the highlighting should be removed (if the text does not match anymore) and in there it's probably not restoring the range properly. Still a bug but at least it should not affect comment highlighting.

src/content.js
peyerluk: It took me a while to understand what the current code in range-save-restore tried to do... I left it mostly in place, added a comment and just fixed the bug 😎

peyerluk:
Ok, I think this is ready. Do you want to have another look before we merge this?

src/cursor.js
```diff
@@ -138,6 +138,7 @@ export default class Cursor {
     this.range.insertNode(element)
     this.range.setStartAfter(preceedingElement)
     this.range.setEndAfter(preceedingElement)
+    this.host.normalize() // mend text nodes
```
marcbachmann, line 141: Is normalize safe to use at this point? Does it maintain the cursor position correctly?
E.g. if the inserted element has a text length of 1, the cursor might jump by 1 character.

peyerluk: I tested it and it seemed to work as intended.

marcbachmann (commented):
Works well. There's at least one new bug with the whitespace highlighting when wrapping the selection in a bold tag. Somehow other highlighting isn't affected, so this is a bug of an edge case we could live with.

image

peyerluk (approved):

peyerluk:
> Add the cursor after the word, add a <space> and <delete> it again. The cursor then jumps before happy.

✅ This is fixed now as well.

peyerluk:
> There's at least one new bug with the whitespace highlighting when wrapping the selection in a bold tag

✅ Fixed.

PR #401, opened 2023-08-31
Support block-based plainText option

dfreier:
Adds an option to configure plain text behavior per block.
- Commands like `toggleBold` won't work on selections within such a block
- Keyboard shortcuts for formatting won't work

To enable the behavior use:
`editable.add('.example', {plainText: true}` or `editable.enable('.example', {plainText: true}`

The normalize option can still be passed as second argument, or it can be included in the options parameter:
`editable.enable('.example', true)` or `editable.enable('.example', {normalize: true}`

### DIFF
--- a/src/block.js
+++ b/src/block.js
@@ -7,11 +7,12 @@ const state = {}
 export const next = getSibling('nextElementSibling')
 export const previous = getSibling('previousElementSibling')
 
-export function init (elem, {normalize, shouldSpellcheck}) {
+export function init (elem, {normalize, plainText, shouldSpellcheck}) {
   setBlockId(elem)
 
   elem.setAttribute('contenteditable', true)
   elem.setAttribute('spellcheck', Boolean(shouldSpellcheck))
+  elem.setAttribute('data-plaintext', Boolean(plainText))
 
   elem.classList.remove(config.editableDisabledClass)
   elem.classList.add(config.editableClass)
@@ -23,6 +24,7 @@ export function init (elem, {normalize, shouldSpellcheck}) {
 export function disable (elem) {
   elem.removeAttribute('contenteditable')
   elem.removeAttribute('spellcheck')
+  elem.removeAttribute('data-plaintext')
 
   setState(elem, undefined)
 
@@ -30,6 +32,9 @@ export function disable (elem) {
   elem.classList.add(config.editableDisabledClass)
 }
 
+export function isPlainTextBlock (elem) {
+  return elem.getAttribute('data-plaintext') === 'true'
+}
 
 export function setBlockId (elem) {
   if (!elem.hasAttribute('data-editable')) {
--- a/src/clipboard.js
+++ b/src/clipboard.js
@@ -2,8 +2,9 @@ import config from './config'
 import * as string from './util/string'
 import * as nodeType from './node-type'
 import * as quotes from './quotes'
+import {isPlainTextBlock} from './block'
 
-let allowedElements, requiredAttributes, transformElements, blockLevelElements, replaceQuotes
+let allowedElements, allowedPlainTextElements, requiredAttributes, transformElements, blockLevelElements, replaceQuotes
 let splitIntoBlocks, blacklistedElements
 const whitespaceOnly = /^\s*$/
 const blockPlaceholder = '<!-- BLOCK -->'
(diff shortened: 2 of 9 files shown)


src/selection.js
```diff
@@ -144,6 +145,9 @@ export default class Selection extends Cursor {
 
   // toggle('<em>')
   toggle (elem) {
+    const elems = content.getTagsByNameAndAttributes(this.host, this.range, elem)
```
dfreier, line 148: Maybe these checks could even be moved to `content.js`, wdyt?

peyerluk: What function would you add to content?

And I think I don't quite understand the use case here. What do we check here? In what circumstance can you toggle an element in a plainText block?

dfreier: There is a `toggleTag()` function in the content.js file. I thought about moving the guarding code over there because then it would not on a selection. The host element (which is the "block") is passed as an argument. In my understanding it would never make sense to toggle a tag on a block which was enabled with plain text option.

peyerluk: >  In my understanding it would never make sense to toggle ON a tag on a block which was enabled with plain text option. But there is no reason why it shouldn't be allowed to toggle OFF a tag in a plain text block.

I would simply disable the option for both cases.

While it may make sense to be able to toggle it off, it would most likely still be weird for a user as you can't undo what you've done. From a UX perspective it's often problematic to allow one-way manipulations. Mostly it's better to simply disable a functionality if it's not fully functional.

peyerluk: Also the code is simpler :)

src/selection.js
```diff
@@ -338,4 +343,9 @@ export default class Selection extends Cursor {
     this.range = content.deleteCharacter(this.host, this.range, character)
     this.setSelection()
   }
+
+  isPlainTextBlock () {
+    const {plainText = false} = block.getState(this.host) ?? {}
```
peyerluk, line 348: It might be easier if we do the normalisation of the state in the `getState()` method. At the moment we don't need a generic state object so we can say it to be an object with a `plainText` property.
And that way we don't need to do the normalisation everywhere the `getState()` method is called.

src/clipboard.js
```diff
@@ -55,16 +58,17 @@ export function paste (element, cursor, clipboardContent) {
  * @param {DOM node} A container where the pasted content is located.
  * @returns {Array of Strings} An array of cleaned innerHTML like strings.
  */
-export function parseContent (element) {
+export function parseContent (element, plainText = false) {
   // Filter pasted content
-  return filterHtmlElements(element)
+  return filterHtmlElements(element, plainText)
```
peyerluk, line 63: The pasting with plainText should also have a few tests.

I suspect you filter the tags out too early and the splitting into blocks doesn't work anymore. 

And instead of using `content.getPlainText(elem)` I think you should be able to just use an empty array for `allowedElements`.

src/content.js
```diff
@@ -22,6 +22,43 @@ export function tidyHtml (element) {
...
+
+  if (element.nodeType === nodeType.elementNode) {
+    let result = ''
+
+    for (const childNode of element.childNodes) {
+      let childText = getPlainText(childNode)
+        .replace(/\s*\n+\s*/g, ' ') // collapse newlines within text
+
+      // collapse leading whitespace
+      if (/^\s+\S+/.test(childText)) {
```
peyerluk, line 38: Maybe we should separate the whitespace normalisation from the tag filtering? Also we have some existing methods for whitespace filtering both in `string.js` as well as in this file. E.g. `removeWhitespaces()`. A thing to note is that when removing whitespaces it can leave empty text nodes which can make the whole thing a bit more complicated if one wants to leave the text nodes in a clean state after cleaning. If memory serves right this is what `removeWhitespaces()` does for trimming in the start and end.

src/block.js
```diff
@@ -17,14 +17,17 @@ export function init (elem, {normalize, shouldSpellcheck}) {
   elem.classList.add(config.editableClass)
 
   if (normalize) content.tidyHtml(elem)
+  if (plainText) content.normalizePlainText(elem)
```
peyerluk, line 20: I think you don't use `getPlainText()` in `extractContent()` yet. Is this an oversight or on purpose?

I thin it would make sense to normalize on init like you do here and every time the framework reads the content into its model where `extractContent()` is used.

peyerluk (changes requested):
A few things to go over. If you'd like we can also do a call to discuss or look at some stuff.

peyerluk (approved):

PR #421, opened 2023-10-20
Remove bowser module exposed on `Editable.browser`

marcbachmann:
### Changelog
- 💥 Remove bowser module exposed on `Editable.browser`


peyerluk (approved):
👏 ~15Kbyte replaced with 3 lines of code

PR #428, opened 2023-11-17
Handle nodes and text when trimming whitespace from a selection

ajwild:
Relations:
  - Issues:
    - Fixes https://github.com/livingdocsIO/livingdocs-bugs/issues/5083
    - Fixes https://github.com/livingdocsIO/livingdocs-bugs/issues/4895

# Motivation

We are trying to trim whitespace characters around selections for certain tags such as links, bold, italic etc. For links we add and remove span elements to save the range when the focus is lost, so we're actually dealing with text nodes for the offsets and not character offsets. For bold and italic it could be text node or character offsets that we're dealing with in the range. The code in master always assumes the offsets are characters.

# Changelog

- 🐞 Handle nodes and text when trimming whitespace from a selection
- 🐞 Don't wrap a collapsed selection with tags

### DIFF
--- a/src/selection.js
+++ b/src/selection.js
@@ -156,6 +156,7 @@ export default class Selection extends Cursor {
   // e.g. toggle('<em>')
   toggle (elem) {
     if (block.isPlainTextBlock(this.host)) return
+    if (this.range.collapsed) return
     elem = this.adoptElement(elem)
     this.range = content.toggleTag(this.host, this.range, elem)
     this.setVisibleSelection()
@@ -320,6 +321,7 @@ export default class Selection extends Cursor {
   // remove first.
   forceWrap (elem) {
     if (block.isPlainTextBlock(this.host)) return
+    if (this.range.collapsed) return
     elem = this.adoptElement(elem)
     this.range = content.forceWrap(this.host, this.range, elem)
     this.setVisibleSelection()
--- a/src/selection.js
+++ b/src/selection.js
@@ -5,7 +5,12 @@ import * as block from './block'
 import config from './config'
 import highlightSupport from './highlight-support'
 import highlightText from './highlight-text'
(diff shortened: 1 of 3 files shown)


ajwild:
@meinradandermatt @JSchenk8 please can you test this and make sure it fixes the reported bugs. It would also be good to test it does not break any other text selection logic. I'll try and add a couple of tests tomorrow for the bug report issues.

src/selection.js
```diff
@@ -352,3 +372,92 @@ export default class Selection extends Cursor {
...
+  }
+
+  // *** Start hack ***
+  // Pass the root so that the iterator can traverse to siblings
+  const iterator = new NodeIterator(root)
+  // Set the position to the node which is selected
+  iterator.nextNode = startContainer
+  // Iterate once to avoid returning self
+  iterator.getNextTextNode()
+  // *** End hack ***
```
ajwild, line 399: @peyerluk please could you take a look at this "hack". In order for this to work I need to set the root element in the `NodeIterator` as the range `commonAncestorContainer`, but the first node should be the range `startContainer`. Should I modify `NodeIterator` to accept different values for both, or keep the "hack" as it is?

Also, it always return the `startContainer` when I call `getNextTextNode()` for the first time (because at this point in the code `startContainer` will be a text node. It looks like this is a deliberate design consideration, so are you happy enough with the "unused" `iterator.getNextTextNode()` call in this code?

peyerluk: For this case it is best you add enough tests for a few different use cases. Then you don't have to worry about this 'hack' too much.

But code like this is an ideal candidate for test driven development.

Also you should move the helper functions out of the selection class.

ajwild: I've added a couple of tests and moved the helper functions.

peyerluk (approved):

PR #433, opened 2023-12-08
Handle nodes and text when trimming whitespace from a selection [3.x]

ajwild:
Backporting https://github.com/livingdocsIO/editable.js/pull/428 to the new 3.x release branch.

Fixes: https://github.com/livingdocsIO/livingdocs-bugs/issues/5135

### DIFF
--- a/spec/selection.spec.js
+++ b/spec/selection.spec.js
@@ -433,14 +433,16 @@ describe('Selection', function () {
     })
 
     it('trims a range with special whitespaces', function () {
-      // at the beginning we have U+2002, U+2005 and U+2006 in the end a normal whitespace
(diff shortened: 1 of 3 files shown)


ajwild:
This introduces a selection bug after adding a link using the extended search.

ajwild:
@marcbachmann I think this can be merged now.

3.2.0 has the selection bug which seems to highlight all text on the page (in the iframe) from the third link onwards:
Peek 2024-01-10 18-53

It must get stuck in a selection mode, or have an invalid selection to restore because the link failed to apply to the selection. I've tried to reproduce this with the new branch and with 5.0.2 and they don't have the same problem as the link is always applied correctly.

marcbachmann (approved):

Issue #262, opened 2022-02-17
Missing module '/src/config' when importing.


Rimevel:
I'm working on a project using SvelteKit for reactivity and backend. When I try to import I get:

Cannot find module 'D:\Git\write\node_modules\@livingdocs\editable.js\src\config' imported from D:\Git\write\node_modules\@livingdocs\editable.js\src\core.js
Error [ERR_MODULE_NOT_FOUND]: Cannot find module 'D:\Git\write\node_modules\@livingdocs\editable.js\src\config' imported from D:\Git\write\node_modules\@livingdocs\editable.js\src\core.js
    at finalizeResolution (internal/modules/esm/resolve.js:276:11)
    at moduleResolve (internal/modules/esm/resolve.js:699:10)
    at Loader.defaultResolve [as _resolve] (internal/modules/esm/resolve.js:810:11)
    at Loader.resolve (internal/modules/esm/loader.js:86:40)
    at Loader.getModuleJob (internal/modules/esm/loader.js:230:28)
    at ModuleWrap.<anonymous> (internal/modules/esm/module_job.js:56:40)
    at link (internal/modules/esm/module_job.js:55:36)

Rimevel:
Also when running it through Vite i sometimes get:
The requested module '/node_modules/.vite/@livingdocs_editable_js.js?t=1645129462300&v=b65e2d94' does not provide an export named 'default'
SyntaxError: The requested module '/node_modules/.vite/@livingdocs_editable_js.js?t=1645129462300&v=b65e2d94' does not provide an export named 'default'

marcbachmann:
I think that error is misleading. The issue is most likely `rangy` that only exports using UMD. https://github.com/livingdocsIO/editable.js/blob/master/src/core.js#L1

I've tried to migrate it somewhen to https://github.com/notjosh/rangy, but failed as not all modules were migrated back then. At some time we should try to migrate everything to the native range api. There aren't many reasons anymore to use that third party library.

Could you try importing dist/editable.js directly? Additionally to that we could generate a `dist/editable.esm.js` or something similar

Rimevel:
If I do import the dist version I get this:

ReferenceError: self is not defined
    at file:///D:/Git/Slashwrite/node_modules/@livingdocs/editable.js/dist/editable.js:1:198
    at ModuleJob.run (internal/modules/esm/module_job.js:152:23)
    at async Loader.import (internal/modules/esm/loader.js:166:24)
[...]

PR #441, opened 2024-01-18
Explicitly use file extensions with imports

marcbachmann:
### Changelog
- 🐞 Explicitly use file extensions with imports. This should fix builds with other tooling than webpack. See #262


robinbisping (approved):
Looks good to me

Issue #488, opened 2024-05-27
First `change` event has old element instance


NelsonBrandao:
The first `change` event triggered has an old version of the element. This results on the change being missed. Further `change` events appear to be working correctly.

This can be seen on demo provided by this package:

https://github.com/livingdocsIO/editable.js/assets/1036806/96267ce7-fc79-42d0-8ae3-0ed7941b9712

The first time we type `1` the output below doesn't get updated.

marcbachmann:
Thanks for reporting this.
It's caused by some backwards compatibilty support of older browsers and we can remove this by now.

I also just ran into the same issue when running e2e tests.

PR #510, opened 2024-07-22
Remove input event backwards compatibility

marcbachmann:
Relations:
- Fixes: #488 

### Changelog
- 🐛 Remove backwards compatibility for missing input events.
  The logic causes issues as it loses the first change/input event.


ajwild (approved):
Thank you! This fixes a bug where the first input event was not detected in Firefox when a component was already highlighted on canvas load. The input event has also had a good level of support for the last 4+ years: https://caniuse.com/input-event.

PR #580, opened 2025-02-18
Smart quotes

eileenoo:
Relations:

- Basecamp Issue: [Link](https://3.basecamp.com/4910529/buckets/39227833/documents/8027430845)
- Related PR's: [Server PR ](https://github.com/livingdocsIO/livingdocs-server/pull/7695)
- Related PR's: [Editor UI PR ](https://github.com/livingdocsIO/livingdocs-editor/pull/9545)

# Motivation

Enhance text consistency and readability by automatically formatting quotation marks as users type. This ensures they match the configured preferred style.

# How to test

- You need to have `livingdocs-framework`, `livingdocs-server`, `livingdocs-editor` and this branch of `editable.js` locally available
- In [`livingdocs-framework`](https://github.com/livingdocsIO/livingdocs-framework)'s `package.json`, point the `@livingdocs/editable.js` dependency to this local version of `editable.js` and then install dependencies
- Have this PR of [`livingdocs-server`](https://github.com/livingdocsIO/livingdocs-server/pull/7695) checked out and do the same as above for the `@livingdocs/framework` dependency and then install dependencies
- Have this PR of [`livingdocs-editor`](https://github.com/livingdocsIO/livingdocs-editor/pull/9545) checked out and do the same for the `@livingdocs/framework` dependency and then install dependencies
- Start [`livingdocs-server`](https://github.com/livingdocsIO/livingdocs-server/pull/7695) and  [`livingdocs-editor`](https://github.com/livingdocsIO/livingdocs-editor/pull/9545)

# Known Edge Cases

<list or text>

- Finding last, unclosed, single opening quote when typing single closing quote and other single closing quote comes before. E.g. `‹Hello ‹stranger››`
- Return to last cursor position, when no input was done -> keyboard jump 'left' / 'right'
- Smart Quotes is not applied, when Paragraph is switched, before timeout is finished
- Different Range container due to bold/italic/link single closing quote is not applied
- Quickly typing multiple quotes after each other only transforms first typed quote. Due the offset being 0

### DIFF
--- a/src/smartQuotes.js
+++ b/src/smartQuotes.js
@@ -44,9 +44,13 @@ export const applySmartQuotes = (range, config, char, target, cursorOffset) => {
     return
   }
 
+  const {quotes, singleQuotes} = config
+  if (char === quotes[0] || char === quotes[1] || char === singleQuotes[0] || char === singleQuotes[1]) {
+    return
+  }
+
   const offset = range.startOffset
   const textArr = [...range.startContainer.textContent]
-  const {quotes, singleQuotes} = config
   let newTextNode
 
   if (shouldBeClosingQuote(textArr, offset - 2)) {
--- a/src/smartQuotes.js
+++ b/src/smartQuotes.js
@@ -1,15 +1,43 @@
 const isValidQuotePairConfig = (quotePair) => Array.isArray(quotePair) && quotePair.length === 2
 
-export const shouldApplySmartQuotes = (config) => {
+export const shouldApplySmartQuotes = (config, target) => {
   const {smartQuotes, quotes, singleQuotes} = config
-  return !!smartQuotes && isValidQuotePairConfig(quotes) && isValidQuotePairConfig(singleQuotes)
+  return !!smartQuotes && isValidQuotePairConfig(quotes) && isValidQuotePairConfig(singleQuotes) && target.isContentEditable
 }
 
-export const isQuote = (char) => /^[‘’‹›‚'«»"“”„]$/.test(char)
-export const isDoubleQuote = (char) => /^[«»"“”„]$/.test(char)
-export const isSingleQuote = (char) => /^[‘’‹›‚']$/.test(char)
+// export const isQuote = (char) => /^[‘’‹›‚'«»"“”„]$/.test(char)
+const isDoubleQuote = (char) => /^[«»"“”„]$/.test(char)
+const isSingleQuote = (char) => /^[‘’‹›‚']$/.test(char)
 
 // Test: '>', ' ', no space & all kinds of dashes dash
-export const isOpeningQuote = (text, indexCharBefore) => indexCharBefore < 0 || /\s|[>\-–—]/.test(text[indexCharBefore])
+const isOpeningQuote = (text, indexCharBefore) => indexCharBefore < 0 || /\s|[>\-–—]/.test(text[indexCharBefore])
+const isClosingQuote = (text, indexCharBefore) => text[indexCharBefore] !== ' '
 
-export const isClosingQuote = (text, indexCharBefore) => text[indexCharBefore] !== ' '
+const applySmartQuote = (textArr, index, target, quoteType) => {
+  if (index >= 0 && index < textArr.length) {
+    textArr[index] = quoteType
+    target.innerText = textArr.join('')
+  }
+}
+
+export const applySmartQuotes = (config, char, wholeText, offset, target, resetCursor) => {
+  const isCharSingleQuote = isSingleQuote(char)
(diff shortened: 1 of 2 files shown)


src/smartQuotes.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+  for (let i = offset - 1; i >= 0; i--) {
+    if (isSingleQuote(textArr[i]) && (!isApostrophe(singleOpeningQuote) && !isApostrophe(textArr[i]))) {
+      return textArr[i] === singleOpeningQuote
+    }
+  }
+  return false
+}
+
+export const applySmartQuotes = (range, config, char, target, carretOffset) => {
```
ajwild, line 39: Please could you change this to `cursorOffset`. We tend to use cursor instead of caret throughout the code.

src/smartQuotes.js
```diff
@@ -0,0 +1,79 @@
...
+  const isCharSingleQuote = isSingleQuote(char)
+  const isCharDoubleQuote = isDoubleQuote(char)
+
+  if (!isCharDoubleQuote && !isCharSingleQuote) {
+    return
+  }
+
+  const offset = range.startOffset
+  const textArr = [...range.startContainer.textContent]
+  const {quotes, singleQuotes} = config
```
ajwild, line 49: This function could also return early if the `char` already matches the expected quote.

src/smartQuotes.js
```diff
@@ -0,0 +1,79 @@
...
+  let newTextNode
+
+  if (shouldBeClosingQuote(textArr, offset - 2)) {
+    if (isCharSingleQuote) {
+      // Don't transform apostrophes
+      if (hasCharAfter(textArr, offset)) {
+        return
+      }
+      // Don't transform single-quote if there is no respective single-opening-quote
+      if (!hasSingleOpeningQuote(textArr, offset, singleQuotes[0])) {
```
ajwild, line 59: It seems like it's missing a `hasDoubleOpeningQuote` check when `isCharDoubleQuote` is true.

src/smartQuotes.js
```diff
@@ -0,0 +1,79 @@
...
+  const {smartQuotes, quotes, singleQuotes} = config
+  return !!smartQuotes && isValidQuotePairConfig(quotes) && isValidQuotePairConfig(singleQuotes) && target.isContentEditable
+}
+
+const isDoubleQuote = (char) => /^[«»"“”„]$/.test(char)
+const isSingleQuote = (char) => /^[‘’‹›‚']$/.test(char)
+const isApostrophe = (char) => /^[’']$/.test(char)
+const isWhitespace = (char) => /^\s$/.test(char)
+
+const shouldBeOpeningQuote = (text, indexCharBefore) => indexCharBefore < 0 || /\s|[>\-–—]/.test(text[indexCharBefore])
```
ajwild, line 13: I'd make another little helper function for this regular expression. Also, is an opening quote really expected immediately after a hyphen/dash etc?

src/smartQuotes.js
eileenoo: The single quotes are checked because we have apostrophes which we don't want to transform. 
That's not the case for double quotes. They are always transformed.

src/smartQuotes.js
eileenoo: I will add it :)

src/smartQuotes.js
eileenoo: Yes, I will change it :)

src/smartQuotes.js
eileenoo: I can extract the regex as helper function :) The dash-logic is analog to the checks in quotes.js and it works the same way in the MACOS smart quotes

src/smartQuotes.js
ajwild: Ok, I see why now. I would double check whether the Mac Smart Quotes feature transforms a closing quote without an opening one earlier in the string. It still seems like it would be a valid check to have.

eileenoo: It does :) 

https://github.com/user-attachments/assets/7b27375a-c72b-4703-be7c-6e3ce90d2033

src/smartQuotes.js
eileenoo: Added like to: 
```js const {quotes, singleQuotes} = config
  if (char === quotes[0] || char === quotes[1] || char === singleQuotes[0] || char === singleQuotes[1]) {
    return
  }```

src/smartQuotes.js
eileenoo: Added the helper function :)

ajwild (approved):

PR #584, opened 2025-02-27
Follow up smart quotes

eileenoo:
### Changelog

- Changed delay from 500ms to 300ms 
- Added special case, where single quote character is transformed into a single opening quote, if the character before is a double quote 
- Added tests for smartQuote helper-functions and replaceQuote functions


spec/smartQuotes.spec.js
```diff
@@ -0,0 +1,116 @@
...
+    expect(invalidNodeValue).to.equal(null)
+  })
+
+  it('should return null if range is empty', () => {
+    const range = createRangeWithText('')
+    const replacedTextNode = replaceQuote(range, 0, '`')
+    expect(replacedTextNode).to.equal(null)
+  })
+
+  it('should insert quote at the end, if index is out of bounce', () => {
```
ajwild, line 111: "out of bounds". Or "outside range".

ajwild (approved):
The timing works well for me.

PR #596, opened 2025-04-03
Remove zero width space when getting text range for comment highlights

ajwild:
Relations:
- Issue: Fixes https://github.com/livingdocsIO/livingdocs-bugs/issues/5048

# Motivation

When creating comments in Livingdocs it uses the selected text to create a highlight. If the text selection contains a zero width non-breaking space, which is removed when the component is blurred, then the highlight never computes the correct start and end offset for the range because the text string no longer matches.

# Changelog

- 🐞 Do not let highlight exceed text range
- 🐞 Remove zero width space to make selection more accurate, because it will be removed on component blur via extractContent


src/highlight-support.js
```diff
@@ -45,8 +45,9 @@ const highlightSupport = {
       this.win
     )
 
-    const actualStartIndex = startIndex
-    const actualEndIndex = endIndex
+    // Do not let highlight exceed text range
+    const actualStartIndex = Math.max(startIndex, 0)
+    const actualEndIndex = Math.min(endIndex, blockText.length)
```
ajwild, line 50: These values could also be clamped, so that the start cannot be greater than one character before the end of the string, and the end cannot be less than 1.

src/highlight-support.js
```diff
@@ -45,8 +45,9 @@ const highlightSupport = {
       this.win
     )
 
-    const actualStartIndex = startIndex
-    const actualEndIndex = endIndex
+    // Do not let highlight exceed text range - it should also be at least 1 character long
+    const actualStartIndex = Math.min(Math.max(startIndex, 0), blockText.length - 1)
+    const actualEndIndex = Math.min(Math.max(endIndex, 1), blockText.length)
```
ajwild, line 50: I want to take another look at this before merging.

src/highlight-support.js
```diff
@@ -45,8 +45,14 @@ const highlightSupport = {
...
     )
 
-    const actualStartIndex = startIndex
-    const actualEndIndex = endIndex
+    // Do not let highlight exceed text range
+    const actualStartIndex = Math.min(Math.max(startIndex, 0), blockText.length - 1)
+    const actualEndIndex = Math.min(Math.max(endIndex, 1), blockText.length)
+
+    // If text is provided then validate that it matches
+    if (text && text !== blockText.slice(actualStartIndex, actualEndIndex)) {
```
ajwild, line 53: The tests are failing because this doesn't handle line breaks etc.

